PR #15, opened 2015-01-06
Add and configure Rollbar support

catsby:
First pass at adding Rollbar to the project


broker/redis_test.go
```diff
@@ -70,6 +72,8 @@ func (s *BrokerSuite) TestRedisSubscribe(c *C) {
 	ch, _ := s.broker.Subscribe()
 	defer s.broker.Unsubscribe(ch)
 	s.broker.Publish([]byte("busl"))
+	// sleep 1 second to avoid race condition on Travis :(
+	time.Sleep(1 * time.Second)
```
tt: I don't understand why this is necessary. Why is an empty string written to the channel?

tt:
Looks good to me.

broker/redis_test.go
catsby: I don't understand either. Think I should ping Travis team about it? Seems like a Redis thing on their end, though I don't fully understand it. I *thought* it could be explained by the go 1.4 and EOF returns, but looking at the commit history here, I switched back to 1.4 and it still occurred without the sleep. I'm at a loss.

tt: Yeah, that seems reasonable. I'm not able to reproduce this locally. Can you?

Unrelated: Did you conclude that the EOF behavior changed? I thought it had always worked like that.

catsby: I don't believe I've been able to reproduce this locally. 
Sadly, it passes locally regardless if the `readLen` block is before or after the switch on `err`. 
I think that's a clue...

PR #16, opened 2015-01-12
Integrate braintree/manners, supports graceful shutdown

catsby:
Use github.com/braintree/manners. 
Code lifted from that other project. 

Ignore the flappy Redis test, it is out to get me

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"log"
 	"net/http"
-	"syscall"
 	"time"
 
 	"github.com/braintree/manners"
@@ -171,7 +170,7 @@ func addDefaultHeaders(fn http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
-func Start() {
+func Start(port string, shutdown <-chan struct{}) {
 	p := pat.New()
 
 	p.GetFunc("/health", addDefaultHeaders(health))
@@ -181,9 +180,9 @@ func Start() {
 
 	http.HandleFunc("/", enforceHTTPS(p.ServeHTTP))
 
-	go listenForShutdown(util.AwaitSignals(syscall.SIGURG))
+	go listenForShutdown(shutdown)
 
-	if err := gracefulServer.ListenAndServe(":"+*util.HttpPort, nil); err != nil {
+	if err := gracefulServer.ListenAndServe(":"+port, nil); err != nil {
 		panic(err)
 	}
 }
--- a/server/server.go
+++ b/server/server.go
@@ -20,8 +20,8 @@ var gracefulServer *manners.GracefulServer
 
 func init() {
 	gracefulServer = manners.NewServer()
-	gracefulServer.InnerServer.ReadTimeout = time.Hour
-	gracefulServer.InnerServer.WriteTimeout = time.Hour
+	gracefulServer.InnerServer.ReadTimeout = *util.HttpReadTimeout
+	gracefulServer.InnerServer.WriteTimeout = *util.HttpWriteTimeout
 }
 
 func mkstream(w http.ResponseWriter, _ *http.Request) {
--- a/server/server.go
+++ b/server/server.go
@@ -6,14 +6,24 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"syscall"
 	"time"
 
+	"github.com/braintree/manners"
 	"github.com/cyberdelia/pat"
 	"github.com/heroku/busl/assets"
 	"github.com/heroku/busl/broker"
 	"github.com/heroku/busl/util"
 	"github.com/heroku/rollbar"
 )
+
+var gracefulServer *manners.GracefulServer
+
+func init() {
+	gracefulServer = manners.NewServer()
+	gracefulServer.InnerServer.ReadTimeout = time.Hour
+	gracefulServer.InnerServer.WriteTimeout = time.Hour
+}
 
 func mkstream(w http.ResponseWriter, _ *http.Request) {
 	registrar := broker.NewRedisRegistrar()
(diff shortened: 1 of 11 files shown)


broker/redis_test.go
```diff
@@ -71,6 +72,7 @@ func (s *BrokerSuite) TestRedisSubscribe(c *C) {
 	ch, _ := s.broker.Subscribe()
 	defer s.broker.Unsubscribe(ch)
 	s.broker.Publish([]byte("busl"))
+	time.Sleep(1 * time.Second)
```
ryanbrainard: this will still probably work with something `< 1 * time.Second`. Try `1 * time.Nanosecond` to keep the tests fast.

server/server.go
```diff
@@ -58,6 +68,14 @@ func pub(w http.ResponseWriter, r *http.Request) {
 	for {
 		readLen, err := bodyBuffer.Read(buffer)
 
+		if readLen > 0 {
```
ryanbrainard: why did this move up?

broker/redis_test.go
catsby: I've actually got it failing with the sleep as is, so I'm not sure anything will help there. Our friends @ Travis sent a pull request (https://github.com/heroku/busl/pull/17) to try their new infrastructure, I'm going to test that out.

server/server.go
```diff
@@ -171,7 +181,17 @@ func Start() {
 
 	http.HandleFunc("/", enforceHTTPS(p.ServeHTTP))
 
-	if err := http.ListenAndServe(":"+*util.HttpPort, nil); err != nil {
+	go listenForShutdown(util.AwaitSignals(syscall.SIGURG))
```
ryanbrainard: not sure what you guys are doing for busl, but in endosome all the OS stuff (e.g. signal handling) was kept in the `main` with just a channel being passed into `server`. This should make a cleaner separation and for better testing.

server/server.go
catsby: Originally on a suspicion, but I found this:

> When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. It may return the (non-nil) error from the same call or return the error (and n == 0) from a subsequent call. An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF regardless.

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.


- http://golang.org/pkg/io/#Reader

The previous behavior (before I changed the case statement as well here https://github.com/heroku/busl/commit/21c383c053e3772c15512188d618b7c8314df3a0#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bL58) would treat `io.EOF` as an error and not use `readLen`, which as I read the docs, is incorrect.

If I've interpreted that wrong, I'm all ears :)

server/server.go
catsby: I originally had that, but I moved it here so I didn't change the `start()` signature. I didn't really consider the separation of concerns (by keeping it in main), which makes sense in hindsight; I'll patch it to match Endosome.

catsby: Hey @ryanbrainard can you take a look at https://github.com/heroku/busl/commit/c973e730d5508f9a1a880741d758b2a05d0f9c84 ? I changed it to be more like Endosome, but not exactly. Specifically, I don't convert `*util.HttpPort` to an integer and pass it in, only to convert that integer back into a string for the `ListenAndServe` call

server/server.go
```diff
@@ -15,6 +16,14 @@ import (
 	"github.com/heroku/rollbar"
 )
 
+var gracefulServer *manners.GracefulServer
+
+func init() {
+	gracefulServer = manners.NewServer()
+	gracefulServer.InnerServer.ReadTimeout = time.Hour
```
tt: Is this a sane value?

server/server.go
tt: It sounds right to me.

server/server.go
tt: I've always found it funky that `Start` doesn't return when it's started. Shouldn't it rather wrap the graceful server and offer a `Stop`?

The main function then becomes:

    server.Start(...)
    <-signal
    server.Stop()

server/server.go
catsby: Shamefully lifted from Endosome. Might not be a suitable value for this :frowning:

server/server.go
ryanbrainard: IRC, `ListenAndServe` blocks, so `Start` would not return to await the signal

tt: It's especially nice because in tests, you can `defer server.Stop()` and not working about spawning the server in a goroutine.

tt: It does block, but the server could just spawn the goroutine. I guess renaming from `Start` to something like `ListenAndWait` would work too.

server/server.go
naaman: This should probably be shorter, but beware that clients may need to retry publishing or subscribing. Currently, there aren't any clients doing that for publishing (not sure about subscribing).

naaman: Also, this should probably be a config value if for no other reason than testing.

server/server.go
naaman: *that feeling when you put something in a specific place after hours of fighting with the code and can't remember why*

server/server.go
naaman: Also, is there a reason this needs to be in init?

server/server.go
```diff
@@ -15,6 +16,14 @@ import (
 	"github.com/heroku/rollbar"
 )
 
+var gracefulServer *manners.GracefulServer
```
naaman: Is there a reason this is global? I'd suggest moving this all into the server startup so that you can test it.

server/server.go
naaman: :+1: from reading the docs.

naaman:
Aside from the nitpicks, :+1: from me

PR #17, opened 2015-01-13
lets try the new container setup

joshk:


catsby:
LGTM; trying it out to see if I can get the test to flip-flop

catsby:
Merged; I wasn't able to reproduce the flip-flop failure after ~7 tries, so I think we're good here. Thanks!

PR #19, opened 2015-01-22
Go.uuid instead of own version

cyx:
Throwing out there if it makes sense


tt:
I'm not a huge fan of that library (or any of the libraries; I have my own which I never released because I thought it was silly with a ton of very basic UUID libraries). It doesn't consider that reading might fail, so you could get an empty UUID.

cyx:
Yep, makes sense.

catsby:
revisiting: what about https://code.google.com/p/go-uuid ?

cyx:
Yeah based on some googling some people complain it's not RFC compliant --
but if we don't care about
that I'm fine using that.

PR #20, opened 2015-01-23
Sketch of doing http.status log lines

cyx:
Was weighing between this approach and just manually doing log.SCountOne for each of the three methods (pub / sub / mkstream)

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -6,9 +6,9 @@ import (
 	"github.com/heroku/busl/util"
 )
 
-func enforceHTTPS(ƒ http.HandlerFunc) http.HandlerFunc {
+func enforceHTTPS(fn http.HandlerFunc) http.HandlerFunc {
 	if !*util.EnforceHTTPS {
-		return ƒ
+		return fn
 	}
 
 	return func(w http.ResponseWriter, r *http.Request) {
@@ -21,7 +21,7 @@ func enforceHTTPS(ƒ http.HandlerFunc) http.HandlerFunc {
 			return
 		}
 
-		ƒ(w, r)
+		fn(w, r)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


server/middleware.go
```diff
@@ -24,3 +24,11 @@ func enforceHTTPS(ƒ http.HandlerFunc) http.HandlerFunc {
 		ƒ(w, r)
 	}
 }
+
+func logRequest(fn http.HandlerFunc) http.HandlerFunc {
```
tt: Perhaps use `ƒ` for consistency?

cyx: Ok, mostly just wondering if we should just go fn instead (linux's support
for special characters leaves something to be desired...)

I can always copy & paste it though :-)

util/response_logger.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (l *responseLogger) Header() http.Header {
+	return l.w.Header()
+}
+
+func (l *responseLogger) Write(b []byte) (int, error) {
+	if l.status == 0 {
+		// The status will be StatusOK if WriteHeader has not been called yet
+		l.status = http.StatusOK
```
tt: Why not initialize with this value to avoid this check?

server/middleware.go
tt: I'm fine changing to something that's easier to type.

tt:
Looks good to me except for a couple very minor details.

util/response_logger.go
```diff
@@ -0,0 +1,45 @@
...
+	l.w.WriteHeader(s)
+	l.status = s
+}
+
+func (l *responseLogger) WriteLog() {
+	maskedStatus := strconv.Itoa(l.status/100) + "xx"
+
+	// TODO: possibly use heroku/slog for writing the
+	// structured logs instead
+	log.Printf("count#http.status.%s=1", maskedStatus)
```
catsby: can we include the request id in the header, if present?

server/middleware.go
catsby: We see `ƒ` in Endosome in a place or two. I would prefer `fn` personally...

cyx:
Closing in favor of #23

PR #21, opened 2015-01-23
Log request-id

cyx:
Making this separate to localize the conversation to this PR.

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -29,6 +29,24 @@ func logRequest(fn http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		logger := util.NewResponseLogger(w)
 		fn(logger, r)
-		logger.WriteLog()
+		logger.WriteLog(requestId(r))
 	}
+}
+
+func requestId(r *http.Request) (id string) {
+	if id = r.Header.Get("Request-Id"); id == "" {
+		id = r.Header.Get("X-Request-Id")
+	}
+
+	if id == "" {
+		// In the event of a rare case where uuid
+		// generation fails, it's probably more
+		// desirable to continue as is with an empty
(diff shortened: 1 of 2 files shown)


cyx:
Closing in favor of #23

catsby:
Can we change `maskedStatus := strconv.Itoa(l.status/100) + "xx"` to log the full status? I think it's useful

catsby:
Other than small comment, :+1:

cyx:
Something worth pointing out: I did add the full status as a separate
key-value pair

catsby:
Oh, I missed that. Why the redundancy then? Tools like Librato (for our use) will aggregate that, yeah?

I'm OK either way, as long as the full status is *somewhere* :smile:

cyx:
Redundancy: mostly I'm just gonna point to endosome. I'm still in C&P mode
for these logging alerts as there's really no document establishing all
rules

catsby:
no worries, :+1:

server/middleware.go
```diff
@@ -29,6 +29,24 @@ func logRequest(fn http.HandlerFunc) http.HandlerFunc {
...
 	}
 }
+
+func requestId(r *http.Request) (id string) {
+	if id = r.Header.Get("Request-Id"); id == "" {
+		id = r.Header.Get("X-Request-Id")
+	}
+
+	if id == "" {
+		// In the event of a rare case where uuid
```
tt: Is it? I don't know, but if /dev/random breaks; it would probably make other things fail too.

cyx: Hmm you mean having an empty request ID should prevent the service
altogether from working?

cyx: I guess that was my thought because at the point that `requestId` will be called, the request is considered to be done, and we're only logging the line.

PR #25, opened 2015-01-26
Implement http.Flusher and http.CloseNotifier

cyx:
- without http.CloseNotifier we get a panic
- without http.Flusher every sub request ends up saying
  that streaming is unsupported


tt:
Could we add a unit test to ensure we don't forget an implementation again?

cyx:
Sure will do that.

PR #27, opened 2015-02-06
Log HTTP Port

naaman:
When busl starts, it uses `$PORT` for listening to HTTP requests.
`$PORT` can be any number and is sometimes opaque. This change
provides a simple logging mechanism to see which port busl used for
HTTP requests.


cyx:
:+1:

catsby:
looks-good-to-me-ship-it

PR #28, opened 2015-02-11
io.Copy friendly spike

cyx:
Wanted to create this to initiate some conversation.

Some things that I'm not too happy with currently:

1) The error handling in publishOn in conjunction with `Write`
   Not sure if this is on purpose that we silently ignore
   the EXEC error instead of failing or not, but then rely
   on the `PUBLISH` call (but also silently ignore the error
   return value there)

2) Potentially making msgBroker more of an io.Reader / io.Writer
   and dropping some of the function calls (if they are deemed
   unnecessary)

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -5,7 +5,7 @@ import (
 )
 
 type Broker interface {
-	Publish(msg []byte)
+	Write(msg []byte) (int, error)
 	Subscribe() (chan []byte, error)
 	Unsubscribe(ch chan []byte)
 }
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -1,11 +1,13 @@
 package broker
 
 import (
+	"io"
+
 	"github.com/heroku/busl/util"
 )
 
 type Broker interface {
-	Write(msg []byte) (int, error)
+	io.Writer
 	Subscribe() (chan []byte, error)
 	Unsubscribe(ch chan []byte)
 }
--- a/broker/redis.go
+++ b/broker/redis.go
@@ -192,7 +192,16 @@ func (b *RedisBroker) Publish(msg []byte) {
 	b.publishOn(msg)
 }
 
-func (b *RedisBroker) publishOn(msg []byte) {
+func (b *RedisBroker) Write(msg []byte) (n int, err error) {
+	err = b.publishOn(msg)
+
+	if err != nil {
+		return 0, err
+	}
+	return len(msg), nil
+}
+
+func (b *RedisBroker) publishOn(msg []byte) error {
 	conn := redisPool.Get()
 	defer conn.Close()
 
@@ -204,10 +213,11 @@ func (b *RedisBroker) publishOn(msg []byte) {
 	appendedVals, err := redis.Values(conn.Do("EXEC"))
 	if err != nil {
 		util.CountWithData("RedisBroker.publishOn.error", 1, "error=%s", err)
+		return err
 	}
 	appendedLen := appendedVals[0].(int64)
 
-	conn.Send("PUBLISH", b.channel.id(), appendedLen)
+	return conn.Send("PUBLISH", b.channel.id(), appendedLen)
 }
 
 func (b *RedisBroker) replay(ch chan []byte) (err error) {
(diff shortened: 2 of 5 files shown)


cyx:
@naaman would appreciate to hear your thoughts since you're the father of busl :-)

broker/broker.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 type Broker interface {
-	Publish(msg []byte)
+	Write(msg []byte) (int, error)
```
tt: `io.Writer` instead.

broker/redis.go
```diff
@@ -204,10 +209,11 @@ func (b *RedisBroker) publishOn(msg []byte) {
 	appendedVals, err := redis.Values(conn.Do("EXEC"))
 	if err != nil {
 		util.CountWithData("RedisBroker.publishOn.error", 1, "error=%s", err)
+		return err
 	}
 	appendedLen := appendedVals[0].(int64)
 
-	conn.Send("PUBLISH", b.channel.id(), appendedLen)
+	return conn.Send("PUBLISH", b.channel.id(), appendedLen)
```
tt: It's likely not a big deal, but this is a change in behavior. Before it was resilient to one-off failures, although the stream would have been incomplete.

broker/broker.go
naaman: @tt I'm confused by your comment, the new signature should satisfy [`io.Writer.Write`](http://golang.org/pkg/io/#Writer). Are you suggesting an embedded type for implementations of `Broker`?

tt: type Broker interface {
    	io.Writer
    	...
    }

naaman: :+1:

broker/redis.go
naaman: Agreed. This will be a p9X problem. It'd probably be good to check how frequently `RedisBroker.publishOn.error` occurs in production before making this change.

cyx: We can probably keep the behavior and silently return nil for the error (even if there was one) just to satisfy the Writer interface.

If you think that's better, I'm :+1: (was mostly unsure regarding the behavior that we want)

broker/redis.go
```diff
@@ -188,11 +188,16 @@ func (b *RedisBroker) UnsubscribeAll() {
 	conn.Do("SETEX", b.channel.doneId(), redisChannelExpire, []byte{1})
 }
 
-func (b *RedisBroker) Publish(msg []byte) {
-	b.publishOn(msg)
+func (b *RedisBroker) Write(msg []byte) (n int, err error) {
+	err = b.publishOn(msg)
```
naaman: There's a missing condition of the `Write` contract in this implementation. From `io` docs:

> Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. Write must return a non-nil error if it returns n < len(p). Write must not modify the slice data, even temporarily.

The missing condition is when you have a partial `Write`. AFAIK, you can't get that from redis. But, it does make me wonder if retries should be considered as an approximation.

naaman:
:+1: to the general approach. I'd suggest implementing `Subscribe` as an `io.Reader` as well (probably in a separate PR). That's something I wanted to clean up, but never got a chance to.

tt:
It's funny you should say that. I suggested it won't be possible to convert it into a reader without making the design worse. Because there in theory can be multiple subscribers per channel, we will have to keep a buffer per reader to keep using a single subscription to the underlying channel. Thoughts?

naaman:
Implementing `Read`, in my view, is a convenience for `Broker` clients. I made the suggestion because I felt it's nice to have a common interface to use with things like `io.Copy`. I don't think it should replace `Subscribe`. The underlying channel should probably be exposed in some way. e.g. For `sub`, you'll still likely need access to the channel to handle the `select` loop.

broker/redis.go
cyx: On the other hand, we've never seen the error happen yet.

https://metrics.librato.com/metrics/RedisBroker.publishOn.error?duration=2419200

PR #29, opened 2015-02-12
Initial spike for sse.Stream idea

cyx:
Was looking at using an existing eventSource library,
but nothing seemed to fit (at least in my view).

### DIFF
--- a/sse/sse.go
+++ b/sse/sse.go
@@ -1,0 +1,50 @@
+package sse
+
+import (
+	"bytes"
+	"fmt"
+)
+
+const (
+	id   = "id: %d\n"
+	data = "data: %s\n"
+)
+
+var (
+	newline = []byte{'\n'}
+)
+
+// Usage:
+//
+//     in := make(chan []byte, 10)
+//     out := make(chan []byte, 10)
+//     go sse.Transform(in, out)
+//
+//     in <- "hello"
+//     <-out // id: 5\ndata: hello\n\n
+//
+func Transform(offset int, in, out chan []byte) {
+	for {
+		msg, msgOk := <-in
+
+		if msgOk {
+			out <- format(offset, msg)
+			offset += len(msg)
+		} else {
+			close(out)
+			return
+		}
+	}
+}
+
+func format(pos int, msg []byte) []byte {
+	buf := bytes.NewBufferString(fmt.Sprintf(id, pos+len(msg)))
+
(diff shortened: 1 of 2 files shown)


sse/sse.go
```diff
@@ -0,0 +1,58 @@
...
+func Stream(offset int, in, out chan []byte) {
+	pos := 0
+
+	for {
+		msg, msgOk := <-in
+
+		if msgOk {
+			length := len(msg)
+
+			if pos >= offset {
```
tt: This isn't efficient. This means we'll still have to stream everything from Redis. That may be several megabytes of logs to just receive the last few bytes.

sse/sse_test.go
```diff
@@ -0,0 +1,91 @@
...
+	input = []string{
+		"hello",
+		"world",
+		"!",
+		"hello\nworld",
+		"hello\n\nworld",
+	}
+
+	output = []string{
+		"id: 0\ndata: hello\n\n",
```
tt: This is incorrect. The ID should match the offset after the data is received or a reconnect will result in the consumer receiving this chunk again.

cyx: :+1:

sse/sse.go
cyx: 2 thoughts for this:

1) We can alter the broker side to accept an offset
2) Right now the stream when they reconnect will basically be one key -- but yes we can always amend the broker side to do offset in conjunction with GETRANGE.

In hindsight wanted to build something less complected, but if we want more efficiency it will have to know some of the internals of the other packages.

cyx: Ah yes -- scratch that. sse can have less responsibility and assume the channel already starts where we want it, the position just being initialized to the offset that's passed.

The heavy lifting will be done by b.Subscribe then. :+1: ?

naaman:
@cyx It may be interesting to note busl was originally implemented (never committed) using SSE. SSE was abandoned because busl was intended to serve as a general purpose stream. In other words, you should be able to stream a binary file, such as a `tar.gz`, using a fairly dumb client, such as curl. This can serve a number of purposes: file uploads, direct communication with a dyno, etc. Moving to SSE would require a client to be smarter about the data being streamed. That said, the general purpose stream requirement may no longer be necessary and SSE might serve some more useful purpose.

tt:
We keep serving both and won't change the publisher side; this is just a convenience for text streams.

PR #30, opened 2015-02-13
Subscribe with offset

cyx:


broker/broker.go
```diff
@@ -8,7 +8,7 @@ import (
 
 type Broker interface {
 	io.Writer
-	Subscribe() (chan []byte, error)
+	Subscribe(offset int64) (chan []byte, error)
```
tt: Unless a stream can really be more than ~2 GiB, `int32`?

broker/redis.go
```diff
@@ -117,6 +118,7 @@ func (b *RedisBroker) Subscribe() (ch chan []byte, err error) {
 	defer b.mutex.Unlock()
 	ch = make(chan []byte, msgBuf)
 	b.subscribers[ch] = true
+	b.position = offset
```
tt: A broker might have multiple subscribers, so this is not safe.

broker/broker.go
cyx: yeah, was mostly playing with the existing position type, but if you think it's safe, yes can change that.

tt:
Looks good to me.

PR #31, opened 2015-02-18
Panic happens if tried concurrently

cyx:
With this fix we ensure that the closed channel
shouldn't be getting any more data.

### DIFF
--- a/broker/redis.go
+++ b/broker/redis.go
@@ -158,11 +158,13 @@ func (b *RedisBroker) redisSubscribe(ch chan []byte, offset int64) {
 			}
 		case redis.Subscription:
 			if msg.Kind == "punsubscribe" || msg.Kind == "unsubscribe" {
-				subscSlice, _ := b.getRange(offset, []byte("-1"))
-				ch <- subscSlice
+				if _, ok := b.subscribers[ch]; ok {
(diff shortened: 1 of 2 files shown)


broker/redis.go
```diff
@@ -150,11 +150,13 @@ func (b *RedisBroker) redisSubscribe(ch chan []byte) {
 			}
 		case redis.Subscription:
 			if msg.Kind == "punsubscribe" || msg.Kind == "unsubscribe" {
-				subscSlice := b.getRange([]byte("-1"))
-				ch <- subscSlice
+				if b.subscribers[ch] {
```
tt: This may seem irrelevant, but I think `ok, _ := b.subscribers[ch]; ok` is clearer. It's not defined what the value means.

broker/redis.go
```diff
@@ -150,11 +150,13 @@ func (b *RedisBroker) redisSubscribe(ch chan []byte) {
...
+					subscSlice := b.getRange([]byte("-1"))
+					ch <- subscSlice
```
tt: Couldn't there still be a race condition?

    select {
    	case ch <- subscSlice:
    	default:
    }

Wouldn't this code solve it?

PR #32, opened 2015-02-18
Add reader sketch

cyx:
Just an idea, there is a way to avoid channels and the like, but this introduces the least amount of changes while still riding the same broker implementation.

### DIFF
--- a/broker/reader.go
+++ b/broker/reader.go
@@ -1,0 +1,41 @@
+package broker
+
+import (
+	"io"
+
+	"github.com/heroku/busl/util"
+)
+
+type Reader struct {
+	broker *RedisBroker
+	ch     chan []byte
+}
+
+func NewReader(uuid util.UUID) (*Reader, error) {
+	broker := NewRedisBroker(uuid)
+	ch, err := broker.Subscribe()
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &Reader{broker: broker, ch: ch}, nil
+}
+
+func (r *Reader) Read(p []byte) (n int, err error) {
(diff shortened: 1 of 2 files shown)


broker/reader.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (r *Reader) Read(p []byte) (n int, err error) {
+	msg, ok := <-r.ch
+
+	if !ok {
+		err = io.EOF
+	}
+
+	if n = len(msg); n > 0 {
```
tt: What if the message is larger than `p`?

broker/reader.go
```diff
@@ -0,0 +1,41 @@
+package broker
+
+import (
+	"io"
+
+	"github.com/heroku/busl/util"
+)
+
+type Reader struct {
```
tt: This can be unexported.

tt:
This does give convenience if you, say, want to use `io.Copy`, but I still wonder how applicable this is to our known use cases. For the chunked encoding, we could do just copy, but then we'd miss the heartbeats. For Server-Sent Events, I think pushes are definitely better than polling. It seems we'd end up having a lot more code without any cleaner usage. Perhaps you can spike the intended usage to demonstrate that it would be cleaner?

cyx:
Yes I can add to this spike a short intended usage.

For the heartbeats was thinking of creating a KeepAliveReader of sort to abstract that part, which can accept the null byte (and we provide :keepalive in the case of an event-stream request).

cyx:
I'll leave this open for now but I'm tempted to just close it :-)

PR #33, opened 2015-02-18
Make event source work

cyx:
Quick spike. Open to suggestions for creating struct abstractions / interfaces to make everything less hacky :-)

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -6,12 +6,14 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/braintree/manners"
 	"github.com/cyberdelia/pat"
 	"github.com/heroku/busl/assets"
 	"github.com/heroku/busl/broker"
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -99,6 +107,17 @@ func sub(w http.ResponseWriter, r *http.Request) {
...
 
+	keepalive := util.GetNullByte()
+	if r.Header.Get("Accept") == "text/event-stream" {
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.Header().Set("Cache-Control", "no-cache")
+
+		out := make(chan []byte, len(ch))
+		go sse.Transform(offset, ch, out)
+		ch = out
+		keepalive = []byte(":keepalive\n")
```
tt: Shouldn't this be a semicolon?

tt:
Looks good to me.

server/server.go
cyx: You mean

    :keepalive

Should be

    :;

?

PR #34, opened 2015-02-18
Add basic makefile

cyx:


cyx:
Had this uncommitted for a while, think it's good to have.

PR #35, opened 2015-02-19
Revert the original behavior of not throwing 5xx

cyx:
Previously we just logged an unexpected EOF. Doing that again here.


tt:
Looks good to me.

PR #36, opened 2015-02-25
Sse fixes

cyx:
Realized was adding more newlines that what was in the input. Added regression testing on the server SSE interface as well.


tt:
I'm not sure I follow this change. The idea was to allow incomplete lines; the mandatory line break shouldn't be considered content.

tt:
We chatted about this and it seems the fix is incorrect.

The format requires a string without a line break to be represented as `data: foo\n\n`. The two line break characters are part of the format. A string with a line break would be `data: foo\ndata:\n\n`.

@appleton, are you seeing something else?

appleton:
The only time I see additional newlines is right at the start (for the app I'm testing anyway). Here's an example of the complete output:

```
id: 1
data: 
data: 

id: 29
data: -----> Node.js app detected
data: 

id: 86
data: 
data: -----> Reading application state
data:        package.json...
data: 

id: 112
data:        build directory...
data: 

id: 138
data:        cache directory...
data: 

id: 170
data:        environment variables...
data: 

id: 359
data: 
data:        Node engine:         0.10.x
data:        Npm engine:          unspecified
data:        Start mechanism:     npm start
data:        node_modules source: package.json
data:        node_modules cached: true
data: 
data: 

id: 506
data:        NPM_CONFIG_PRODUCTION=true
data:        NODE_MODULES_CACHE=true
data: 
data: -----> Installing binaries
data:        Resolving node version 0.10.x via semver.io...
data: 

id: 556
data:        Downloading and installing node 0.10.36...
data: 

id: 627
data:        Using default npm version: 1.4.28
data: 
data: -----> Building dependencies
data: 

id: 668
data:        Restoring node modules from cache
data: 

id: 703
data:        Pruning unused dependencies
data: 

id: 737
data:        Installing any new modules
data: 

id: 769
data: 
data: -----> Checking startup method
data: 

id: 950
data:        No Procfile; Adding 'web: npm start' to new Procfile
data: 
data: -----> Finalizing build
data:        Creating runtime environment
data:        Exporting binary paths
data:        Cleaning npm artifacts
data: 

id: 981
data:        Cleaning previous cache
data: 

id: 1022
data:        Caching results for future builds
data: 

id: 1048
data: 
data: -----> Build succeeded!
data: 
data: 

id: 1220
data:        node-js-sample@0.1.0 /tmp/build_bb6ae88bd648876c0c9e7df4d0308523/appleton-sample-node-app-38b87b6
data:        â”œâ”€â”€ ejs@1.0.0
data:        â””â”€â”€ express@4.11.2
data:        
data: 

id: 1253
data: -----> Discovering process types
data: 

id: 1314
data:        Procfile declares types -> web
data: 
data: -----> Compressing... 

id: 1326
data: done, 5.7MB
data: 

id: 1346
data: -----> Launching... 

id: 1427
data: done, v192
data:        https://pure-anchorage-9855.herokuapp.com/ deployed to Heroku
data: 
data: 

id: 1427
data: 

```

As @tt says the double newlines in the middle of the stream are correctly signifying a string with a linebreak.

cyx:
Got it yeah right now I omit the newlines at the start as well so all
should be good.

On that note, the SSE 204 termination is done in the io branch.

cyx:
This was fixed in the io branch

PR #37, opened 2015-03-02
Io

cyx:
Sketch of pure reader / write approaches as opposed to channels

@naaman ^

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -1,16 +1,8 @@
 package broker
 
 import (
-	"io"
-
 	"github.com/heroku/busl/util"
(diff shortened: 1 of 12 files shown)


naaman:
@cyx This might take some time to do a proper review. However, the use of connections stuck out. I'm not sure we should let the consumer of a reader decide when a redis pool connection should be released. The pooled connection should always be released and the `Close()` implementation would just be a no-op (if it's even necessary).

cyx:
Got it; I guess with the none-channel approach there's no other way to work around that.

PR #40, opened 2015-03-11
Authenticater

cyx:


server/middleware.go
```diff
@@ -27,6 +28,18 @@ func enforceHTTPS(fn http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
+func auth(fn http.HandlerFunc) http.HandlerFunc {
+	if *util.Creds == "" {
+		return fn
+	}
+
+	if auth, err := authenticater.NewBasicAuthFromString(*util.Creds); err != nil {
```
freeformz: This makes a new authenticator for each request, which is overkill.

Make one, re-use it use the exported WrapAuth function in place of the auth function here.

cyx: Tried it locally and it seems the `auth` function is only called once per invocation in `app()`

tt: I think the point is that the same instance can cover multiple handlers. I don't think that's an optimization we need to care about, though, and we only have a single authenticated endpoint.

freeformz: Ahh true. I did a quick read and assume a different pattern here w/o really thinking much. Sorry.

Upon re-reading ... you will create 2 of these, which is fine.

I would probably put this ins server.go, in app(), and remove this method entirely.

server/server.go
```diff
@@ -161,8 +161,8 @@ func app() http.Handler {
 	p := pat.New()
 
 	p.GetFunc("/health", addDefaultHeaders(health))
-	p.PostFunc("/streams", addDefaultHeaders(mkstream))
-	p.PostFunc("/streams/:uuid", addDefaultHeaders(pub))
+	p.PostFunc("/streams", auth(addDefaultHeaders(mkstream)))
+	p.PostFunc("/streams/:uuid", auth(addDefaultHeaders(pub)))
```
freeformz: Why not:

```go
auth, err := NewBasicAuthFromString(*util.Creds);
if err != nil {
  return err
}

p.PostFunc("/streams", authenticater.WrapAuth(auth,addDefaultHeaders(mkstream)))
p.PostFunc("/streams/:uuid", authenticater.WrapAuth(auth,addDefaultHeaders(pub)))
```

I would probably go one further and move the creation of auth outside of the call to app() and pass it in so different Authenticaters can be injected. (PS: why isn't this called newApp() or some such)? This is useful for testing where auth doesn't matter for the test.

PR #41, opened 2015-03-11
Don't authenticate POST /streams/<id> requests

cyx:
- Only authenticate the stream creation endpoint (`POST /streams`)
- Return 410 if an attempt to publish to a non registered channel happens

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -152,16 +152,21 @@ func (s *HttpServerSuite) TestBinaryPubSub(c *C) {
 	pubRequest := newRequestFromReader("POST", sf("/streams/%s", streamId), bodyCloser)
 	pubResponse := CloseNotifierRecorder{httptest.NewRecorder(), make(chan bool, 1)}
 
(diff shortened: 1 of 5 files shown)


server/server_test.go
```diff
@@ -64,13 +64,13 @@ func (s *HttpServerSuite) Test410(c *C) {
...
 
-func (s *HttpServerSuite) TestPub(c *C) {
+func (s *HttpServerSuite) TestPubNotRegistered(c *C) {
 	request := newRequest("POST", "/streams/1234", "")
 	response := httptest.NewRecorder()
 
 	pub(response, request)
 
-	c.Assert(response.Code, Equals, http.StatusOK)
+	c.Assert(response.Code, Equals, http.StatusGone)
```
tt: Shouldn't this be `http.StatusNotFound`? We don't know that it ever existed.

tt:
Looks good to me but the title is confusing...

PR #42, opened 2015-03-12
Docker Compose

cyx:
- Configure docker-compose.yml
- Configure Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ MAINTAINER Heroku Build & Packaging Team <build-and-packaging@heroku.com>
 COPY . /app
 WORKDIR /app
 ENV HOME /app
+ENV PATH $PATH:$HOME/bin
 RUN mkdir -p /var/lib/buildpack /var/cache/buildpack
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+web:
+  command: busl
+  build: .
+  links:
+   - redis
+  ports:
+    - "5000:5000"
+  environment:
+    PORT: 5000
```
tt: I know this is the Heroku default, but how about using 8080 instead?

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM heroku/cedar:14
+MAINTAINER Heroku Build & Packaging Team <build-and-packaging@heroku.com>
+COPY . /app
+WORKDIR /app
+ENV HOME /app
+ENV PATH $PATH:$HOME/bin
+RUN mkdir -p /var/lib/buildpack /var/cache/buildpack
+RUN git clone --depth 1 https://github.com/heroku/heroku-buildpack-go.git /var/lib/buildpack
```
tt: This buildpack is actually in S3: https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/go.tgz

PR #44, opened 2015-03-28
Storage url

cyx:
- Add concept of storage backend
- Storage backend defined in `STORAGE_BASE_URL` and resolved together with the request URI of the current request (with `/streams/` stripped out)
- Refactor `sub` handler a lot


broker/redis.go
```diff
@@ -73,10 +73,6 @@ func (c channel) wildcardId() string {
 	return string(c) + "*"
 }
 
-func (c channel) uuid() util.UUID {
-	return util.UUID(c)
-}
-
 func (c channel) doneId() string {
 	return string(c) + "done"
```
tt: This leaves a situation where you could use a channel ID with a "done" suffix and get into trouble...

broker/redis_test.go
```diff
@@ -15,12 +15,12 @@ func Test(t *testing.T) { TestingT(t) }
 
 type RegistrarSuite struct {
 	registrar Registrar
-	uuid      u.UUID
+	uuid      string
```
tt: `channel`?

server/keepalivereader.go
```diff
@@ -14,6 +14,7 @@ type payload struct {
 }
 
 type keepAliveReader struct {
+	rd       io.Reader
```
tt: `r`

broker/redis.go
cyx: The old null byte to the rescue then as separator? Good catch

server/keepalivereader.go
cyx: Lol :)

server/middleware.go
```diff
@@ -90,6 +97,83 @@ func offset(r *http.Request) (n int) {
...
+	return int64(n)
+}
+
+// Given URL:
+//   http://build-output.heroku.com/streams/1/2/3?foo=bar
+//
+// Returns:
+//   1/2/3?foo=bar
+func requestURI(r *http.Request) string {
+	return key(r) + "?" + r.URL.RawQuery
```
tt: This will append a question mark even if no query string is provided.

server/middleware.go
```diff
@@ -90,6 +97,83 @@ func offset(r *http.Request) (n int) {
...
+
+	rd, err := broker.NewReader(key(r))
+
+	// Not cached in the broker anymore, try the storage backend as a fallback.
+	if err == broker.ErrNotRegistered {
+		return storage.Get(requestURI(r), offset)
+	}
+
+	if offset > 0 {
+		if seeker, ok := rd.(io.Seeker); ok {
```
tt: Shouldn't this have an else block that will consume those first N bytes?

server/middleware.go
cyx: Yeah thought about the trailing question mark and does not really matter.
Can add an if though :)

server/middleware.go
cyx: For the specific readers in question the assumption is that it should
already be pre- seeked via the range header.

server/server_test.go
```diff
@@ -20,6 +19,7 @@ type HttpServerSuite struct{}
 
 var _ = Suite(&HttpServerSuite{})
 var sf = fmt.Sprintf
+var baseURL = *util.StorageBaseUrl
```
tt: `StorageBaseURL`

storage/blob.go
```diff
@@ -0,0 +1,182 @@
...
+	client := &http.Client{Transport: transport}
+
+	res, err := client.Do(req)
+	if err == nil {
+		switch {
+		case res.StatusCode == 416:
+			err = ErrRange
+		case res.StatusCode >= 500:
+			err = Err5xx
+		case res.StatusCode >= 400:
```
tt: Why group these together? Can we create a type for those we expect and leave everything else?

server/keepalivereader.go
tt: It's idiomatic Go. It's never `rd` in official packages.

storage/blob.go
cyx: 416: user facing
4xx: chan not registered
5xx: retryable

Mostly those scenarios demand inspection when the error happens.

Making an err type increases complexity a lot with type assertions -- tried
it

server/keepalivereader.go
cyx: Rd in stdlib sometimes just not public stuff im meh :)

storage/blob.go
tt: The point is that the range of client errors cover very different scenarios. A bad request means we have to change our client whereas a not found is normal operation. I'd rather handle the ones we care about and let everything else bubble up.

broker/redis.go
ojacobson: we only expect a-zA-Z0-9, dashes, and slashes in the URLs, right?

tt: A prefix/suffix for both channels is probably the best solution.

PR #47, opened 2015-04-29
Delay expiration of streams that are not producing

cyx:
To reproduce:

- You need to have a build which produces output in long
  increments (e.g. every 5 minutes)
- What happens: the channel will show up as `Channel not registered`.


tt:
Looks good to me.

PR #49, opened 2015-05-07
Initial stab at busltee

cyx:


cmd/busltee/main.go
```diff
@@ -0,0 +1,97 @@
+package main
+
+import (
+	"errors"
+	"flag"
```
tt: Can we use https://github.com/ogier/pflag or something else that will give us standard option parsing?

cmd/busltee/main.go
```diff
@@ -0,0 +1,97 @@
...
+		if status, ok := exit.Sys().(syscall.WaitStatus); ok {
+			return status.ExitStatus()
+		}
+	}
+
+	return 0
+}
+
+func main() {
+	url := flag.String("url", "", "busl url")
```
tt: Make this a positional argument instead; the flags are optional, but this is required.

cmd/busltee/main.go
```diff
@@ -0,0 +1,97 @@
...
+	insecure := flag.Bool("insecure", false, "allows insecure SSL connections")
```
tt: `-k` shorthand with proper option parsing?

cmd/busltee/main.go
```diff
@@ -0,0 +1,97 @@
...
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+)
+
+// TODO: Use net/http when this issue has been fixed:
+// @see https://github.com/golang/go/issues/6574
```
tt: Have you been able to reproduce this?

cyx: Yes that was hours of misery

cmd/busltee/main.go
cyx: We don't require them for codon. You can use busltee without a URL

cmd/busltee/main.go
tt: Ha ha; can we override the transport? I feel shelling out has been part of our problems earlier.

cmd/busltee/main.go
tt: Why should you be able to do that?

cmd/busltee/main.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"errors"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+
+	flag "github.com/heroku/busl/Godeps/_workspace/src/github.com/ogier/pflag"
```
tt: Can you `godep -r` so we use the local path? This will make it work with `go get` even when the library changes.

cmd/busltee/main.go
```diff
@@ -0,0 +1,99 @@
...
+		args = append(args, "-k")
+	}
+
+	cmd := exec.Command("curl", args...)
+	cmd.Stdin = stdin
+
+	return cmd.Run()
+}
+
+func run(args []string, writer io.WriteCloser) error {
```
tt: Can we take both `stdout` and `stderr` even though they are the same writer?

cmd/busltee/main.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Wait for http request to complete
+	<-uploaded
+
+	if err != nil {
+		os.Exit(exitStatus(err))
+	}
+}
+
+var usage = "Usage: busltee <url> [-k|--insecure] [--connect-timeout N] -- <command>"
```
tt: This is constant.

cmd/busltee/main.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return 0
+}
+
+func main() {
+	insecure := flag.BoolP("insecure", "k", false, "allows insecure SSL connections")
+	timeout := flag.String("connect-timeout", "5", "max number of seconds to connect to busl URL")
+
+	if flag.Parse(); len(flag.Args()) < 2 {
+		println(usage)
```
tt: `println` is not encouraged.

cmd/busltee/main.go
tt: Also, should the be `io.Closer`? Do we want them closed?

cmd/busltee/main.go
cyx: I only do godep -r. Are you sure godep -r is what you want?

cmd/busltee/main.go
cyx: done (re: separating)

- io.Closer -- the method is the ideal place to do a deferred io.Close on the writer, and moving the call to the close will add more indirection.

cmd/busltee/main.go
tt: Ignore me.

cyx:
Woot! https://github.com/golang/go/commit/e5febf957f5be9d3325c2d851bff6ec7c55e4662

cyx:
@tt @ojacobson any last minute comments? So far so good, been baking solid in prod.

tt:
Considering it's in production, it should probably be merged, but I have one minor comment.

I'd like `ParseFlags` to live in the main package; I think it has a strong relationship with the binary, so I don't want to treat it as reusable.

busltee/runner.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+func Run(url string, args []string, conf *config) (exitCode int) {
+	defer monitor("busltee.busltee", time.Now())
+
+	reader, writer := io.Pipe()
+	done := post(url, reader, conf)
+
+	if err := run(args, writer, writer); err != nil {
+		log.Printf("count#busltee.Run.error=1 error=%v", err.Error())
```
tt: Can we make this `count#busltee.exec.error` for similarity with other metrics? Also, we have some of these tracked already; what are they?

cyx: Just user errors mostly, or compile errors in most

PR #50, opened 2015-05-09
Draft specification for publisher reconnects

cyx:
Draft specification, feel free to scrutinize as much as you can.

### DIFF
--- a/docs/reconnects.md
+++ b/docs/reconnects.md
@@ -1,0 +1,39 @@
+# publisher reconnects brainstorm
+
+In order to make the service 100% more reliable, we need
+to add reconnects to busl.
+
+The following is a draft of the protocol specification:
+
+## Clients
+
+1. Clients keep track of the byte offsets, and send them
+   as `Content-Range: byte=x-y` headers
+2. The client decides how much to buffer, and/or how frequently
+   should the posts happen.
+3. For every chunk that should be sent, the client sends to
+   a new endpoint, which is tentatively defined as `PATCH /streams/<id>`
+4. The client will then have to confirm the `ACK` from the server,
+   and keep retrying the last post until they get an ack. (For cases
+   where the server might be down completely, the client may opt
+   to limit the threshold of retries to a certain number N)
+5. When the client is done, they'll need to ping an endpoint, tentatively
+   defined as `DELETE /streams/<id>` (feel free to suggest a better
+   endpoint for this one).
+
+## Servers
+
+1. Servers will make sure that writes to `PATCH` are stored in the DB,
+   and will then respond with an `ACK`.
+2. If the client sends the same content-range that already exists, the
+   server may opt to disregard the message. If the content-range is an
+   overlap, the server will have to get the delta, and store it
+   accordingly.
+3. If the client sends a content-range that seems to imply that some
+   data was skipped, the server will just append that data (and assume
+   that the client didn't try hard enough).
+4. When the client signals `done`, the server closes the connections to
+   all existing subscribers to the stream. Reconnects to the stream by
+   subscribers with `range` (or `last-event-id` for SSE) that signifies
+   a completed stream will get a `204 No Content` response from this point.
+   


docs/reconnects.md
```diff
@@ -0,0 +1,39 @@
+# publisher reconnects brainstorm
+
+In order to make the service 100% more reliable, we need
+to add reconnects to busl.
+
+The following is a draft of the protocol specification:
+
+## Clients
```
ojacobson: `## Publisher Clients` perhaps? I had a moment of cognitive dissonance trying to understand Content-Range in terms of the way clients handle resumption (especially SSE clients).

docs/reconnects.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. Clients keep track of the byte offsets, and send them
+   as `Content-Range: byte=x-y` headers
+2. The client decides how much to buffer, and/or how frequently
+   should the posts happen.
+3. For every chunk that should be sent, the client sends to
+   a new endpoint, which is tentatively defined as `PATCH /streams/<id>`
```
ojacobson: What's the rationale for a new endpoint per retry?

docs/reconnects.md
```diff
@@ -0,0 +1,39 @@
...
+4. The client will then have to confirm the `ACK` from the server,
+   and keep retrying the last post until they get an ack. (For cases
+   where the server might be down completely, the client may opt
+   to limit the threshold of retries to a certain number N)
+5. When the client is done, they'll need to ping an endpoint, tentatively
```
ojacobson: `DELETE` feels like the wrong verb here, since that's meant to imply that the resource ceases to be available, and that's not what happens. However, I can't think of a better one; if we document the slightly off-spec way `DELETE` is used, I think this is fine.

docs/reconnects.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Clients
+
+1. Clients keep track of the byte offsets, and send them
+   as `Content-Range: byte=x-y` headers
+2. The client decides how much to buffer, and/or how frequently
+   should the posts happen.
+3. For every chunk that should be sent, the client sends to
+   a new endpoint, which is tentatively defined as `PATCH /streams/<id>`
+4. The client will then have to confirm the `ACK` from the server,
```
ojacobson: Is the `ACK` distinct from a 2xx-series response? (`No Content`, perhaps?)

cyx: Just a 2xx in this case, since PATCH is responding to some data posted. Not
sure which 2xx though :)

docs/reconnects.md
```diff
@@ -0,0 +1,39 @@
+# publisher reconnects brainstorm
+
+In order to make the service 100% more reliable, we need
```
tt: 100 % more reliable?

docs/reconnects.md
tt: We might document it so that `DELETE` triggers end of streaming, persists it to the durable storage and expires the read endpoint after some amount of time. It may not be completely intuitive, but think it can work.

docs/reconnects.md
tt: Publishers?

tt:
Perhaps I read this wrong, but this replaces streaming? I think streaming should still be preferred and then we just need to figure out how to negotiate missing content on reconnects. Otherwise subscribers will notice the chunking (because we don't flush regularly) or we'll have very high overhead on each chunk.

tt:
I was thinking that if the connection breaks, the publisher tries to reconnect with a `Content-Range: x-` header. That offset might be a kilobyte earlier than when the connection broke and busl will then be able to accept this and skip the duplicate content or reject the request with an `Accept-Range: x-` header indicating how much content it needs.

cyx:
yes streaming first and reconnects in the form of patches

tt:
Couldn't we resume streaming, though?

cyx:
We can, by breaking existing protocol -- i.e. not killing subscribers when
we receive the closer chunk.

If we do that, we can potentially do some kind of content range based
offsets, but I don't know how that would work -- i.e. if busl breaks mid
stream, how does the client know up to what they should send? Should they
always start from scratch?

Those details I didn't flesh out.

tt:
That's why I proposed that busl can reject and set the `Accept-Range` header so the publisher knows exactly what to resend.

tt:
If the first change we make is making completion explicit, we can deploy everything in a backward compatible way.

dmathieu:
Closing this. We have an internal RFC discussing it.

PR #51, opened 2015-05-15
Read buffer fix

cyx:
A bit overdue. This fixes certain panics that happen if the data > 32768 (the length of the allocated buffer in io.Copy).

### DIFF
--- a/broker/io.go
+++ b/broker/io.go
@@ -55,6 +55,7 @@ type reader struct {
 	replayed bool
 	closed   bool
 	mutex    *sync.Mutex
+	buffered bool
 }
 
 func NewReader(key string) (io.ReadCloser, error) {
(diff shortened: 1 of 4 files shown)


cyx:
@tt +1 ?

cyx:
@ojacobson can I get a +1?

ojacobson:
Aren't you on leave? :D I'm looking, I'm looking

cyx:
I am on "leave"

broker/io.go
```diff
@@ -129,6 +128,10 @@ func (r *reader) replay(p []byte) (n int, err error) {
 			copy(p, buf)
```
ojacobson: I take it this was previously panicking if `buf` was too large after `r.fetch`?

cyx: Hmm yeah the trace includes that line yes :)

ojacobson: I'm convinced, then. +1 from me.

PR #54, opened 2015-05-29
Simplify server_test

cyx:
1. newRequest wrapper has minimal utility
2. sf var was defined but not used anymore
3. CloseNotifier is used but not utilized

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -2,11 +2,9 @@ package server
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


tt:
Less code always looks good to me.

cyx:
Merge? (Also, re: contrib guidelines, I assume we can freely merge given
that this doesn't affect runtime)

tt:
Please do.

PR #56, opened 2015-06-04
Assign upper limit for running child processes

cyx:
As much as we like the solution of enforcing this
on an upper layer, pragmatism dictates that having
this sanity check on busltee itself will reduce
headaches moving forward.


tt:
I'm not a fan of this. It doesn't seem like `busltee` should have a responsibility of terminating ill-behaving child processes.

What if this is either a read time-out or another binary that could be inserted in the chain?

cyx:
We can use this: http://linux.die.net/man/1/timeout

PR #58, opened 2015-06-15
Persist the stream before getting finish ACK

cyx:
For some cases, we don't ever persist whatever we get, and
thus makes it hard to debug what the build was doing before
it hung completely.

This should persist the existing output every X minutes (set
to 5 minutes) so that we can get a better understanding of the
failure conditions for certain cases.


server/server.go
```diff
@@ -90,8 +91,24 @@ func pub(w http.ResponseWriter, r *http.Request) {
...
+				return
+			case <-time.After(*util.StorageInterval):
+				// Asynchronously upload the output to our defined storage backend.
+				go storeOutput(key(r), requestURI(r))
+			}
+		}
+
+	}()
+
+	close(done)
```
tt: When this is closed, the routine will immediately pick up the first case, so there is really no change here.

tt:
I don't think we should write every X minutes. We can write whenever a publisher closes the connection; that will definitely happen at some point.

cyx:
Closing in favor of #59

cyx:
Leaving this open in case #59 fails to do its job.

cyx:
#59 seems to do its job! Closing

server/server.go
```diff
@@ -90,8 +108,7 @@ func pub(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Asynchronously upload the output to our defined storage backend.
-	go storeOutput(key(r), requestURI(r))
+	close(done)
```
freeformz: As discussed in hipchat ... You need to defer this above or you'll leak go routines.

cyx: Thanks!

tt:
Why was this reopened? Can we change to write on publisher disconnects to reduce the number of writes?

cyx:
Mostly reopened to track the fact that the hanging builds still aren't
getting captured. The problem is that the forever hanging doesn't
disconnect.

We can probably do something there and change busltee to disconnect after X
seconds.

tt:
"Forever" is limited to 24 hours by the nature of running inside a dyno.

cyx:
Yep, but 24 hours extension is probably something too risky. You guys can
play with that value though, but my guess is that it might cause huge jumps
in redis usage.

dmathieu:
Do we still need this?
I'm guessing if we've been able to live without for almost a year, it's not so much of an issue?

dmathieu:
Closed this as stale.

PR #61, opened 2015-07-15
Re-add busltee

cyx:


ojacobson:
:+1:

freeformz:
ship it ?

cyx:
Just to note, busltee needs to be compiled against Go master. @freeformz
does this already for most of his stuff but wanted to highlight that.

(Otherwise streams won't be streams, and will be 4096 chunks of data
written out)

freeformz:
@cyx We can likely just use 1.5beta1?

freeformz:
FWIW: which is what I run now locally

cyx:
Yes for sure 1.5beta1 works

PR #62, opened 2015-07-16
Rewrite `busltee ... url -- command` as a wrapper, and add `command | busltee ... url` mode.

ojacobson:
This ensures that Busltee itself is not on the signal path for any commands
it runs. However, this can also cause the started process' output to
continue after the process completes, as the output is relayed through
children which are not terminated immediately.


cyx:
Some golang-isms will be commented on by @freeformz I imagine. Worth asking: what does this look like in bash? :)

busltee/config.go
```diff
@@ -0,0 +1,43 @@
...
+	LogPrefix   string
+	LogFile     string
+	ContentType string
+	Source      Source
+	Program     string
+	Args        []string
+	Self        string
+}
+
+func (self *Config) ExecArgs() []string {
```
freeformz: Why not just use flag?

busltee/config.go
```diff
@@ -0,0 +1,43 @@
+package busltee;
```
freeformz: drop the ;

freeformz:
golint complaints:
```
config.go:7:6: exported type Source should have comment or be unexported
config.go:10:2: exported const Piped should have comment (or a comment on this block) or be unexported
config.go:14:6: exported type Config should have comment or be unexported
config.go:28:1: exported method Config.ExecArgs should have comment or be unexported
config.go:28:1: receiver name should be a reflection of its identity; don't use generic names such as "me", "this", or "self"
exec.go:10:1: exported function Exec should have comment or be unexported
log.go:13:1: exported function OpenLogs should have comment or be unexported
log.go:21:1: exported function CloseLogs should have comment or be unexported
log.go:33:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
pipe.go:8:1: exported function Pipe should have comment or be unexported
```

busltee/exec.go
```diff
@@ -0,0 +1,73 @@
...
+	execErr := syscall.Exec(program, conf.Args, os.Environ())
+	// execErr is almost certainly non-nil here, since Exec doesn't return if
+	// it succeeds. The %v format does the right thing regardless.
+	log.Printf("count#busltee.exec.exec.error=1 error=%v", execErr)
+	return execErr
+}
+
+func startTee(conf *Config, cwd string, stream *os.File) error {
+	// Has to be a native pipe; `io.Pipe` pipes aren't shared between processes.
+	reader, writer, pipeErr := os.Pipe()
```
freeformz: just call it `err`. I would also call the other variables `r` and `w`, but I'm less concerned about those.

busltee/exec.go
```diff
@@ -0,0 +1,73 @@
...
+		Dir: cwd,
+		Env: os.Environ(),
+		Files: []uintptr{
+			reader.Fd(), // stdin
+			stream.Fd(), // stdout
+			os.Stderr.Fd(), // stderr
+		},
+	}
+
+	self, pathErr := exec.LookPath(conf.Self)
```
freeformz: just call it `err`

busltee/exec.go
```diff
@@ -0,0 +1,73 @@
...
+	if pathErr != nil {
+		return pathErr
+	}
+
+	_, forkErr := syscall.ForkExec(self, conf.ExecArgs(), attrs)
```
freeformz: Try
```
if _, err := syscall.ForkExec(self, conf.ExecArgs(), attrs); err != nil {
  return err
}
```

busltee/exec.go
```diff
@@ -0,0 +1,73 @@
...
+	if forkErr != nil {
+		return forkErr
+	}
+
+	dupErr := syscall.Dup2(int(writer.Fd()), int(stream.Fd()))
```
freeformz: Try:
if err := syscall.Dup2(int(writer.Fd()), int(stream.Fd())); err != nil {
  return err
}

busltee/exec.go
```diff
@@ -0,0 +1,73 @@
...
+		return pathErr
+	}
+
+	stdoutTeeErr := startTee(conf, cwd, os.Stdout)
+	if stdoutTeeErr != nil {
+		log.Printf("count#busltee.exec.tee.error=1 error=%v", stdoutTeeErr)
+		return stdoutTeeErr
+	}
+ 
+	stderrTeeErr := startTee(conf, cwd, os.Stderr)
```
freeformz: try:
```
if err := startTee(conf, cwd, os.Stderr); err != nil {
  log.Printf("count#busltee.exec.tee.error=1 error=%v", err)
  return err
}
```

freeformz:
@ojacobson consider changing up the error handling to just be

```
err := ....
if err != nil {
  ...
}
```

or if err or the other returned value(s) aren't used later on try:

```
if err := ....; err != nil {
  ...
}
```

freeformz:
Any particular reason you are passing a *Config around instead of just a Config?

ojacobson:
@freeformz It seems I'd misunderstood Go's scoping rules; I had believed that

```go
a, err := foo()
// ..
b, err := bar()
```

would be illegal due to redeclaring `err`, but not so.

ojacobson:
Golint's happy. I even added the missing comments. :)

ojacobson:
Pointers to `Config`: no particular reason. That was already there. I've rewritten the code to use copies of `Config` throughout, since it's read-only almost everywhere anyways.

busltee/config.go
```diff
@@ -0,0 +1,51 @@
...
+	Source      Source
+	Program     string
+	Args        []string
+	Self        string
+}
+
+// ExecArgs converts its receiver back into an execve-style argument list,
+// usable to launch further busltee instances with the same configuration. The
+// returned list always configures busltee to use the `Piped` source.
+func (conf Config) ExecArgs() []string {
```
freeformz: How about?
```
// ExecArgs returns an  execve-style argument list, usable to launch further busltee
// instances with the same configuration. The returned list always configures busltee
// to use the `Piped` source.
```

busltee/config.go
```diff
@@ -0,0 +1,51 @@
...
+
+// A Source indicates which source busltee will forward to busl.
+type Source int
+
+const (
+	// Piped sources forward stdin to busl.
+	Piped Source = iota
+	// Program sources forward the output of a program to busl.
+	Program
+)
```
freeformz: These are usually done this way to make the iota more prominent

```go
const (
	Piped   Source = iota // Piped sources forward stdin to busl.
	Program               // Program sources forward the output of a program to busl.
)
```

busltee/config.go
```diff
@@ -0,0 +1,51 @@
...
+	Timeout     float64
+	Retry       int
+	URL         string
+	LogPrefix   string
+	LogFile     string
+	ContentType string
+	Source      Source
+	Program     string
+	Args        []string
+	Self        string
```
freeformz: What is this?

busltee/exec.go
```diff
@@ -0,0 +1,75 @@
...
+		return err
+	}
+ 
+	if err := startTee(conf, cwd, os.Stderr); err != nil {
+		log.Printf("count#busltee.exec.tee.error=1 error=%v", err)
+		return err
+	}
+
+	err = syscall.Exec(program, conf.Args, os.Environ())
+	// execErr is almost certainly non-nil here, since Exec doesn't return if
```
freeformz: s/execErr/err

busltee/exec.go
```diff
@@ -0,0 +1,75 @@
...
+	// it succeeds. The %v format does the right thing regardless.
+	log.Printf("count#busltee.exec.exec.error=1 error=%v", err)
```
freeformz: You may want %q in all of these for proper quoting.

busltee/pipe.go
```diff
@@ -0,0 +1,15 @@
+package busltee
```
freeformz: Why a single file for a single Function?

busltee/exec.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// Exec runs the program described in `conf.Program` and `conf.Args` with its 
+// stdout and stderr streams piped through busltee with a similar configuration.
+func Exec(conf Config) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Printf("count#busltee.exec.getwd.error=1 error=%v", err)
+		return err
+	}
```
freeformz: For all of these errors why not just wrap it in an fmt.Errorf and let the caller handle the logging?

freeformz:
One last thing. I don't understand the package layout. the busltee package is sort of a library, sort of not. I would just put all of this in cmd/busltee unless there is a real case for making it a full library.

ojacobson:
@freeformz I've flattened the package structure a bit. Hopefully I've done it sanely: it reads okay, but `package main` in that many files feels odd to me.

PR #66, opened 2016-02-29
Monitor and better handle net timeouts

dmathieu:
A timeout isn't an error we should get notified on rollbar about.


tt:
Looks good to me.

PR #67, opened 2016-02-29
Monitor io.Copy errors in sub requests

dmathieu:


server/server.go
```diff
@@ -108,7 +109,22 @@ func sub(w http.ResponseWriter, r *http.Request) {
 		handleError(w, r, err)
 		return
 	}
-	io.Copy(newWriteFlusher(w), rd)
+	_, err = io.Copy(newWriteFlusher(w), rd)
+
+	if err == io.ErrUnexpectedEOF {
```
tt: Why special case this?

dmathieu: Because we have the same special case in `pub`.

tt: Is there a reason for it to be there? I can't find any data for that metric.

dmathieu: I believe it could happen if we lose the http connection is unexpectedly closed.
That might not happen often, but could during an incident.

tt: Let's not special case it if it isn't causing noise in Rollbar. We should see errors bubble up before determining to silence them.

dmathieu: WFM. I've removed it.

tt:
Looks good to me.

PR #68, opened 2016-03-22
Log requests path with their status

dmathieu:


util/response_logger.go
```diff
@@ -6,13 +6,15 @@ import (
 	"strconv"
 )
 
-func NewResponseLogger(w http.ResponseWriter, requestID string) *responseLogger {
-	return &responseLogger{ResponseWriter: w, status: http.StatusOK, requestID: requestID}
+// NewResponseLogger creates a new logger for http responses
```
tt: "HTTP"

tt:
Nitpick aside, looks good to me.

tt:
This can be another pull request but should we also log the user agent?

dmathieu:
Good idea. I've added it.

PR #69, opened 2016-03-23
Make requests logging closer to the platform router's format

dmathieu:
* Send the count in another log entry.
* Log all the data we can have, so we respect as much as possible the [router log format](https://devcenter.heroku.com/articles/http-routing#heroku-router-log-format).


tt:
Looks good to me.

PR #70, opened 2016-03-25
Upgrade Go

tt:


dmathieu:
How about moving workspace to vendor too?

tt:
@dmathieu, that's what I want; I just split it into multiple pull requests.

PR #74, opened 2016-03-29
Allow passing a request-id to busltee

dmathieu:


busltee/runner.go
```diff
@@ -94,6 +97,10 @@ func streamNoRetry(url string, stdin io.Reader, conf *Config) error {
 	// it from being closed prematurely (and thus allowing writes
 	// on the other end of the pipe to work).
 	req, err := http.NewRequest("POST", url, ioutil.NopCloser(stdin))
+	if conf.RequestID != "" {
+		req.Header.Set("Request-ID", conf.RequestID)
```
tt: `Request-Id` (that's how Go canonicalizes header names).

dmathieu: Sure, though having it has `Request-ID` wouldn't cause any issue.

tt:
Looks good to me.

PR #75, opened 2016-03-30
Move the main command to cmd/busl/

dmathieu:
So we can do `go get github.com/heroku/busl`, and get both busltee and busl.


tt:
I made this yesterday but didn't commit it. Looks good to me.

PR #76, opened 2016-04-06
Don't send invalid utf8 strings

dmathieu:


tt:
Please update Godep and rebase this. Dependencies should only go into the vendor directory now.

tt:
I feel this may be the wrong fix too. Why do busl receive invalid encoded data? Why is Codon ever sending that?

dmathieu:
My bad. I don't know what happened there with Godeps.

dmathieu:
From what I see, codon only forwards the logs.
It doesn't inspect them (we'd probably have encoding errors there otherwise). Which is why I've done that here.

Furthermore, bytes manipulation (and generation of invalid ones) is much easier in go than ruby.

tt:
But why does different tools at random output invalid byte sequences? That doesn't make any sense to me. Could it rather be the heartbeats that interfere with the content?

dmathieu:
This does appear to have been coming from the build: https://github.com/npm/npm/issues/12196
http://status.npmjs.org/incidents/665tfzq37xph

Thanks @hunterloftis

dmathieu:
Closing as stale.

PR #78, opened 2016-05-12
Misc cleanup

dmathieu:
This includes a fairly big cleanup of the packages, basically achieving the following:

* Pass golint.
* No more global flags. Provide options and pass local arguments instead.
* Remove the util package.
* Try to improve lisibility here and there.


dmathieu:
@freeformz: is there anything in here which doesn't look like a go idiom, or that you would do differently?

broker/io.go
```diff
@@ -59,14 +63,15 @@ type reader struct {
 	buffered bool
 }
 
+// NewReader creates a new io read closer
```
freeformz: This documentation doesn't actually explain much.

broker/io_test.go
```diff
@@ -209,15 +209,15 @@ func TestRedisReadFromClosed(t *testing.T) {
 	assert.Equal(t, err, io.EOF)
 
 	// We'll get true here because r.closed is already set
-	assert.True(t, ReaderDone(r))
+	assert.True(t, readerDone(r))
```
freeformz: assert libs aren't idiomatic. I much prefer something like this:

```go
if ! readerDone(r) {
  t.Errorf("readerDone(r) != true. Should be true because <reason>")
}
```

This is just straight forward go, so there is no cognitive overhead wrt what assert.True is doing or what it will produce.

broker/redis.go
```diff
@@ -69,26 +69,29 @@ func (c channel) id() string {
...
+func (c channel) doneID() string {
 	return string(c) + ":done"
 }
 
-func (c channel) killId() string {
+func (c channel) killID() string {
 	return string(c) + ":kill"
 }
 
+// RedisRegistrar is a channel storing data on redis
```
freeformz: If it's not already documented, we may want a doc.go that describes what a "channel" is. I keep reading that and thinking of a go channel, but that's not what it's talking about. Maybe also consider renaming channel to something else to avoid confusion.

logging/main.go
```diff
@@ -1,4 +1,4 @@
-package util
```
freeformz: probably don't want to call this file "main.go". There's nothing wrong with it technically though.

logging/main.go
```diff
@@ -10,7 +10,8 @@ import (
 const prefix = "busl"
 const defaultDomain = "development"
 
-func init() {
+// Configure configures the default logger
+func Configure() {
```
freeformz: What was wrong with init?

cmd/busl/main.go
```diff
@@ -5,21 +5,78 @@ import (
...
+	HTTPPort         = flag.String("httpPort", os.Getenv("PORT"), "HTTP port for the server.")
+	HTTPReadTimeout  = flag.Duration("httpReadTimeout", time.Hour, "Timeout for HTTP request reading")
+	HTTPWriteTimeout = flag.Duration("httpWriteTimeout", time.Hour, "Timeout for HTTP request writing")
+
+	RollbarEnvironment = flag.String("rollbarEnvironment", os.Getenv("ROLLBAR_ENVIRONMENT"), "Rollbar Enviornment for this application (development/staging/production).")
+	RollbarToken       = flag.String("rollbarToken", os.Getenv("ROLLBAR_TOKEN"), "Rollbar Token for sending issues to Rollbar.")
 )
 
 func main() {
+	logging.Configure()
```
freeformz: What is it configuring and how?

logging/main.go
dmathieu: It's loaded very obscurely, by silently importing util. Now we're explicitely saying we want it configured and when.

cmd/busl/main.go
```diff
@@ -5,21 +5,78 @@ import (
...
+	s := server.NewServer(conf)
+	s.ReadTimeout = *HTTPReadTimeout
+	s.WriteTimeout = *HTTPWriteTimeout
+	s.Start(*HTTPPort, awaitSignals(syscall.SIGURG))
+}
+
+func parseFlags() (*server.Config, error) {
+	conf := &server.Config{}
+
+	flag.StringVar(&conf.Credentials, "creds", os.Getenv("CREDS"), "user1:pass1|user2:pass2")
```
freeformz: Why are some of these flags here and some above?

cmd/busl/main.go
dmathieu: It sets the log [prefix](https://golang.org/pkg/log/#SetPrefix) and [flags](https://golang.org/pkg/log/#SetFlags).
Would you have this named differently?

server/server.go
```diff
@@ -1,166 +1,70 @@
...
-	registrar := broker.NewRedisRegistrar()
-
-	if err := registrar.Register(key(r)); err != nil {
-		http.Error(w, "Unable to create stream. Please try again.", http.StatusServiceUnavailable)
-		rollbar.Error(rollbar.ERR, fmt.Errorf("unable to register stream: %#v", err))
-		util.CountWithData("put.create.fail", 1, "error=%s", err)
-		return
+// NewServer creates a new server instance
+func NewServer(config *Config) *Server {
+	s := &Server{
```
freeformz: `return &Server {...}`

server/server.go
```diff
@@ -1,166 +1,70 @@
...
-	}
-
-	netErr, ok := err.(net.Error)
-	if ok && netErr.Timeout() {
-		util.CountWithData("server.pub.read.timeout", 1, "msg=\"%v\"", err.Error())
-		handleError(w, r, netErr)
-		return
-	}
+// Start starts the server instance
+func (s *Server) Start(port string, shutdown <-chan struct{}) {
```
freeformz: This server isn't very composable. Why is it not a http.Handler?

dmathieu: Good idea. I'll refactor it away into one.

freeformz:
I made some comments, but the PR is huge. Any chance to break this up into smaller pieces?

logging/main.go
freeformz: It's still pretty obscure. Your configuring it without passing anything. So how is it configured?

dmathieu:
I've split this into commits showing my iterations process. Splitting this into several PRs without having conflicts would be difficult though.

assets/cat.go
```diff
@@ -1,18 +0,0 @@
-package assets
-
-var HttpCatGone = `░░░░░░░░░░██░░░░░░░░░░██░░░░░░░░
```
naaman: 😢

server/errors.go
```diff
@@ -4,20 +4,36 @@ import (
...
-	"github.com/heroku/busl/assets"
 	"github.com/heroku/busl/broker"
+	"github.com/heroku/busl/logging"
 	"github.com/heroku/busl/storage"
-	"github.com/heroku/busl/util"
 )
 
 var errNoContent = errors.New("No Content")
 
+const asciiGone = `░░░░░░░░░░██░░░░░░░░░░██░░░░░░░░
```
naaman: oh hey, there it is 😸

broker/io_test.go
dmathieu: Maybe. But that'd mean refactoring all tests in the entire project. I think this PR is already big enough :)

server/server.go
dmathieu: I'm keeping this in mind, but fear this might raise the PR's size quite a bit. Since this is how the server currently is in master, I believe we can roll with those changes, and do a proper refactoring of it afterwards.

broker/redis.go
dmathieu: How about renaming "channel" to "feed"?

dmathieu: Then again, I'd suggest doing so in a follow-up PR.

dmathieu:
Note that this PR does look long. Most of it is vendor though.

README.md
```diff
@@ -37,7 +37,6 @@ and then:
 ```sh
 $ godep go install
 $ cp .sample.env .env
-$ export $(cat .env)
```
tt: What's the motivation for replacing this with a library that's also used in production?

broker/io.go
```diff
@@ -6,15 +6,19 @@ import (
...
-	"github.com/heroku/busl/util"
+	"github.com/heroku/busl/logging"
 )
 
 type writer struct {
 	channel channel
 }
 
-var ErrNotRegistered = errors.New("Channel is not registered.")
+// known errors
```
tt: Do we need this comment?

broker/io_test.go
```diff
@@ -9,29 +9,29 @@ import (
...
+	uuid := uuid.NewV4().String()
 	registrar.Register(uuid)
 	r, _ := NewReader(uuid)
 	w, _ := NewWriter(uuid)
 
 	return r, w
 }
 
-func ExamplePubSub() {
+func Example_pub_sub() {
```
tt: Is this really the preferred format?

broker/io_test.go
tt: I agree.

broker/redis.go
tt: I think "channel" is well-established for the publish-subscribe pattern so I would prefer not renaming even if Go also uses that term. "Topic" would probably work too.

tt: I agree that this should be a separate pull request, though.

broker/io_test.go
dmathieu: Yes. golint complains without the underscores.
https://blog.golang.org/examples#TOC_4.

broker/io.go
dmathieu: Yes. golint asks for a comment for the var block.

README.md
dmathieu: Not having to pollute the global dev environment with environment variables, which are very easy to forget when starting work, and won't be removed unless the terminal is closed and reopened afterwards.

broker/redis.go
dmathieu: Topic looks good to me.

PR #79, opened 2016-05-17
Pass golint on all packages

dmathieu:
And remove the `assets` package, as it only holds a logo, which makes sense to be at the top of the project.


broker/io.go
```diff
@@ -13,8 +13,12 @@ type writer struct {
 	channel channel
 }
 
-var ErrNotRegistered = errors.New("Channel is not registered.")
+// known errors
```
tt: Is this required?

tt: I see you responded to that on the other pull request. I think this is really annoying. The comment is not adding any value; it's just making the file a line longer.

util/util.go
```diff
@@ -22,40 +22,17 @@ func NewUUID() (string, error) {
...
 }
 
-type NullByte []byte
-
-func GetNullByte() []byte {
-	return new(NullByte).Get()
-}
-
-func (nb NullByte) Get() []byte {
-	nb = []byte{0}
```
tt: Why was this not initialized with the type, @naaman?

tt:
Looks good to me.

broker/io.go
dmathieu: I agree. But golint is an overall quite useful tool, and ignoring some warnings is really boring.

util/util.go
naaman: 👻

PR #80, opened 2016-05-17
rename the busltee package to publisher

dmathieu:
Having a command and package named busl can be quite misleading as to  where something should be.
This rename makes it clearer this package is the lower-level interface allowing publishing data.


tt:
I think that's actually a Go pattern; that the `cmd/...` package should only contain very little logic and delegate to a package.

The package is not a generic publisher. It's limited to the `tee` like operation required by the command.

dmathieu:
My point here is mostly that the term `busltee` doesn't have any obvious meaning, and it can be hard to guess what this package does at first.
Even though this is not a generic publisher, I believe renaming this package makes it clearer on what it does.

dmathieu:
@tt is there a name you'd better agree with?

dmathieu:
Let's punt on this for now.

PR #81, opened 2016-05-18
Rename the sse package to encoders

dmathieu:
SSE is a non-obvious abreviation.
And we will need to add other encoders to allow ranges with chunked encoding


tt:
Looks good to me.

PR #83, opened 2016-07-12
Use Heroku Redis rather than RedisToGo

gudmundur:
Updates the `app.json` to use Heroku Redis.


dmathieu:
The default plan will be hobby-dev if you don't specify one, and it also means we won't have to update app.json when new plans come out.
Could you remove the plan?

dmathieu:
I've rebased this and add a commit to use the default plan.

dmathieu:
:+1:

PR #84, opened 2016-07-25
Add Access-Control-Expose-Headers.

ojacobson:
I believe the lack of this header is preventing iOS clients from streaming.
Inspecting an iOS dashboard session shows us sending back `Content-Type:
text/event-stream`, but iOS Safari complains that the returned response was
`Content-Type: text/plain`.

I cribbed the actual list from API.

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -64,6 +64,7 @@ func addDefaultHeaders(fn http.HandlerFunc) http.HandlerFunc {
 		}
 		w.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
 		w.Header().Set("Access-Control-Allow-Headers", "Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token")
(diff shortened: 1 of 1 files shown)


server/middleware.go
```diff
@@ -64,6 +64,7 @@ func addDefaultHeaders(fn http.HandlerFunc) http.HandlerFunc {
 		}
 		w.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
 		w.Header().Set("Access-Control-Allow-Headers", "Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token")
+		w.Header().Set("Access-Control-Expose-Headers", "Cache-Control, Content-Language, Content-Type, Expires, Last-Modified")
```
tt: I think it would make sense to omit `Content-Language` (if possible) because we don't support it and really can't (at least assuming this is a general purpose component).

ojacobson: Fixed - we have fairly free reign with this list.

tt:
Looks good to me.

PR #85, opened 2016-07-27
Added charset utf-8 to Content-Type response header. supports iOS

jkutner:


server/middleware.go
```diff
@@ -141,7 +141,7 @@ func newReader(w http.ResponseWriter, r *http.Request) (io.ReadCloser, error) {
 	}
 
 	if r.Header.Get("Accept") == "text/event-stream" {
-		w.Header().Set("Content-Type", "text/event-stream")
+		w.Header().Set("Content-Type", "text/event-stream; charset=utf-8")
```
ojacobson: The only thing I'm not 100% convinced on yet is whether we're actually careful to send UTF-8. (I believe we actually send through arbitrary bytes, as provided by the underlying build output, but don't hold me to that.)

Any thoughts?

tt:
We should test this in all major browsers when it's in staging to make sure it didn't break any other clients.

tt:
In our testing, this didn't change the behavior so I'm closing the pull request.

PR #87, opened 2016-09-06
Setup Makefile tidy/precommit tasks

dmathieu:
And cleanup readme so it matches the command we already have in Makefile.


Makefile
```diff
@@ -1,8 +1,29 @@
...
+travis: tidy test
+
 test: .PHONY
 	env $$(cat .env) go test ./...
 
+setup: hooks tidy
+	cp .env.sample .env
+
+hooks:
+	ln -fs ../../bin/git-pre-commit.sh .git/hooks/pre-commit
```
tt: Is this path accurate?

tt:
Looks good to me if the path is accurate.

Makefile
dmathieu: Yes. that path is relative to where the symbolic link it created.

> ›  git:(master) ls -ll .git/hooks/
total 16
lrwxr-xr-x  1 dmathieu  454177323   27 Sep  6 14:32 pre-commit -> ../../bin/git-pre-commit.sh
-rwxr-xr-x  1 dmathieu  454177323  443 Sep  6 13:47 pre-push

PR #89, opened 2016-09-07
Reconnect: handle ranges for text seeders too

dmathieu:
This adds handling of ranges for non-SSE (plain text) seeders as well, so subscribers can handle reconnections.
This was handled for SSE connections already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,9 @@ busl - the bustle part of hustle.
 
 a simple pubsub service that runs on Heroku.
 
-## usage
+## Usage
+
+### Creating streams
 
 create a stream:
 
@@ -14,6 +16,8 @@ $ export STREAM_ID=$(curl http://localhost:5001/streams -X POST)
 # STREAM_ID=b7e586c8404b74e1805f5a9543bc516f
 ```
 
+### Subscribe
+
 connect a consumer using the stream id:
 
 ```
@@ -21,6 +25,19 @@ $ curl http://localhost:5001/streams/$STREAM_ID
 ...
 ```
 
+#### Disconnections
+
+Subscribers can sometimes get disconnected. If the instance is cycled or deployed for example.
+When this happens, you can specify a `Range` header specifying which where your streaming stopped.
+
+```
+$ curl http://localhost:5001/streams/$STREAM_ID -H "Range: 100-"
+```
+
+SSE connections also handle the `Last-Event-ID` header.
+
+
+### Publish
 in a separate terminal, produce some data using the same stream id...
 
 ```
@@ -29,7 +46,7 @@ $ curl -H "Transfer-Encoding: chunked" http://localhost:5001/streams/$STREAM_ID 
 
 ...and you see the busl.
 
-## setup
+## Setup
 
 to setup to test and run busl, setup [godep](http://godoc.org/github.com/tools/godep)
 and then:
@@ -38,7 +55,7 @@ and then:
 $ make setup
 ```
 
-## test
+## Test
 
 to run tests:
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -5,7 +5,9 @@ busl - the bustle part of hustle.
 
 a simple pubsub service that runs on Heroku.
 
-## usage
+## Usage
+
+## Creating streams
```
dmathieu: :+1: good catch!

README.md
```diff
@@ -14,13 +16,28 @@ $ export STREAM_ID=$(curl http://localhost:5001/streams -X POST)
...
 ...
 ```
 
+#### Disconnections
+
+Subscribers can sometimes get disconnected. If the instance is cycled or deployed for example.
+When this happens, you can specify a `Range` header specifying which where your streaming stopped.
+
+```
+$ curl http://localhost:5001/streams/$STREAM_ID -H "Range: 100-"
```
dmathieu: Yeah, but that's not how the current implementation works for SSE.
https://github.com/heroku/busl/blob/master/server/middleware.go#L77-L79

Looks like changing this would mean changing how [dashboard handles ranges](https://github.com/heroku/dashboard/blob/828223081a8aaf105e1b8c98cba391690b1fbb9b/app/adapters/application.js#L28) (not sure this is actually where it happens. But it seems to be the most likely candidate.

encoders/main.go
```diff
@@ -0,0 +1,6 @@
+package encoders
+
+type Encoder interface {
```
dmathieu: Good call on `ReadSeeker`.

encoders/text_test.go
```diff
@@ -0,0 +1,55 @@
...
+	// passing to LimitReader: comparably similar
+	// to scenario when reading from an http.Response
+	r := strings.NewReader("hello world")
+	r.Seek(10, 0)
+
+	// Use LimitReader to hide the Seeker interface
+	lr := io.LimitReader(r, 11)
+
+	enc := NewTextEncoder(lr)
+	enc.(io.Seeker).Seek(10, 0)
```
dmathieu: Indeed it isn't. :+1:

encoders/text.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func (r *textEncoder) Seek(offset int64, whence int) (n int64, err error) {
+	if seeker, ok := r.Reader.(io.ReadSeeker); ok {
+		r.offset, err = seeker.Seek(offset, whence)
+	} else {
+		// The underlying reader doesn't support seeking, but
+		// we should still update the offset so the IDs will
+		// properly reflect the adjusted offset.
+		r.offset += offset
```
dmathieu: But do we really need to handle it? It's not handled for SSE seekers.
We also don't have any base (it's always the beginning), nor do we have an end or a limit.

encoders/sse_test.go
```diff
@@ -28,16 +28,16 @@ var (
...
 )
 
-func TestNoNewline(t *testing.T) {
-	for _, data := range testdata {
+func TestSSENoNewline(t *testing.T) {
+	for _, data := range testSSEData {
 		r := strings.NewReader(data.input)
 		enc := NewSSEEncoder(r)
-		enc.(io.Seeker).Seek(data.offset, 0)
+		enc.Seek(data.offset, 0)
```
dmathieu: :+1:

encoders/text.go
dmathieu: WFM. I've done that.

README.md
dmathieu: See 59795254781d4c9660d1d97e74457f49801814c3

PR #90, opened 2016-10-25
Remove the mkstream endpoint

dmathieu:
This endpoint has been deprecated for a while, hasn't been called in the past 4 weeks.


tt:
Nice!

PR #91, opened 2016-10-27
Endpoint to know how many bytes have been received

dmathieu:
This introduces a new `HEAD` endpoint which can be used by publishers to know how many bytes have been received and restart streaming at that byte.

I'll set it up in busltee in another PR.


server/endpoints.go
```diff
@@ -71,7 +71,29 @@ func (s *Server) pub(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		if err == broker.ErrClosed {
+			w.WriteHeader(http.StatusGone)
+			return
+		}
+		handleError(w, r, err)
+		return
+	}
+
+	w.Header().Set("Range", fmt.Sprintf("bytes=%d-", length))
```
tt: I think this is improper use of that header. It should probably be `Content-Length` with the size instead.

server/endpoints.go
```diff
@@ -71,7 +71,29 @@ func (s *Server) pub(w http.ResponseWriter, r *http.Request) {
...
+
```
tt: This line is empty.

dmathieu: Line removed.

server/endpoints.go
dmathieu: [Content-Length](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13) is used to send the length of the request body.
It would be good if we were sending the entire output in the request body.

Note that I've reused most of what [google cloud](https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload) does for resumable uploads, hence the `Range` header.

server/endpoints.go
```diff
@@ -71,7 +71,28 @@ func (s *Server) pub(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		handleError(w, r, err)
+		return
+	}
+	defer rd.Close()
+
+	length, err := broker.Len(rd)
+	if err != nil {
+		if err == broker.ErrClosed {
+			w.WriteHeader(http.StatusGone)
```
dmathieu: Google Cloud uses a 308 Resume Incomplete header here.
That header is very confusing though, as it actually has two meanings.
I believe a 410 makes sense too, as it indicates the client needs to back off.

See http://insanecoding.blogspot.fr/2014/02/http-308-incompetence-expected.html?m=1

server/endpoints.go
tt: I think `Content-Length` is appropriate:

> The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.

A `GET` request (assuming a completed stream) would be exactly that length although it may be encoded differently.

server/endpoints.go
```diff
@@ -71,7 +71,28 @@ func (s *Server) pub(w http.ResponseWriter, r *http.Request) {
...
+	rd, err := broker.NewReader(key(r))
+	if err != nil {
+		handleError(w, r, err)
+		return
+	}
+	defer rd.Close()
+
+	length, err := broker.Len(rd)
+	if err != nil {
+		if err == broker.ErrClosed {
```
tt: Can you move this up so it's `if err == broker.ErrClosed { ... } else if err != nil { ... }` instead?

server/endpoints.go
tt: I think it would be cleaner if it failed when trying to write. A `HEAD` request shouldn't really care about that.

server/endpoints.go
dmathieu: All right, `Content-Length` does work for HEAD, as the spec says:

> In the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.

I need to stop looking into how Google does thing.

dmathieu:
We're taking another approach.

PR #93, opened 2016-10-31
Ignore already-sent data when reconnecting a publisher

dmathieu:


broker/io.go
```diff
@@ -252,3 +253,16 @@ func RenewExpiry(rd io.Reader) {
 	conn.Send("EXPIRE", r.channel.id(), redisChannelExpire)
 	conn.Do("EXEC")
 }
+
+// Len returns the length of data already send to the reader
+func Len(rd io.WriteCloser) (int64, error) {
+	conn := redisPool.Get()
+	defer conn.Close()
+
+	strlen, err := redis.Int64(conn.Do("STRLEN", rd.(*writer).channel.id()))
```
tt: The cast to `*writer` isn't great for an exposed method that accepts a `io.WriteCloser`.

dmathieu: Nice catch. I've changed it to make sure we return a valid error when we can't cast.

PR #94, opened 2016-10-31
Busltee reconnect

dmathieu:
Handle publisher reconnections in busltee

### DIFF
--- a/busltee/transport.go
+++ b/busltee/transport.go
@@ -18,6 +18,10 @@ type Transport struct {
 
 func (t *Transport) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	t.body = &bodyReader{req.Body, &bytes.Buffer{}, true}
+	if t.Transport == nil {
+		t.Transport = &http.Transport{}
+	}
+
 	req.Body = t.body
 	req.ContentLength = -1
 	return t.tries(req)
--- a/busltee/transport.go
+++ b/busltee/transport.go
@@ -1,0 +1,71 @@
+package busltee
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
+)
+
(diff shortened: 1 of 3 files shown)


busltee/transport_test.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		expectedBody += "hello world\n"
+		callCount += 1
+
+		if callCount < 4 {
+			server.CloseClientConnections()
+		}
+	})
+	transport := &Transport{
+		Transport:  &http.Transport{},
```
tt: Is this necessary?

dmathieu: Right now, the transport doesn't set a default underlying one. So yes.
I could move that into transport.go though.

dmathieu: Done in 7ff11a6340e10dfad3b18baca29a7f4136894b57

busltee/transport.go
```diff
@@ -0,0 +1,75 @@
...
+
+	req.Body = t.body
+	req.ContentLength = -1
+	return t.tries(req)
+}
+
+func (t *Transport) tries(req *http.Request) (*http.Response, error) {
+	res, err := t.Transport.RoundTrip(req)
+
+	if err != nil {
```
joshwlewis: @dmathieu I toyed around with my own reconnection logic for busl a little while back, and based on my findings, I don't think this will work once busl is behind the heroku router.

The heroku router returns a 503 to the client when the app restarts or if the request is terminated due to inactivity. http only returns errors if it couldn't complete the request; a 500 status code doesn't return an error. I think you'll need to change this conditional to also retry if the response error code is 50x.

dmathieu: @joshwlewis thanks. That's what I discovered yesterday too. I'm on that today.

dmathieu: See #96

PR #97, opened 2016-11-04
Sleep between reconnections, don't store the buffer in memory and make sure the streamer is closed between each connection

dmathieu:
This stores the buffer into a temporary file instead of the memory. Thanks @joshwlewis.
It also fixes an issue, as reading the buffer can't always read everything at the same time, depending on the provided `p` length.

It also introduces sleeping for a second between each reconnect, to avoid exhausting our retries in under a second.

Finally, is moves us to using a different `bodyReader` for each connection, allowing us to make sure we properly close it and older connections don't keep reading on the data.


joshwlewis:
@dmathieu This is working for Heroku CI in staging. Sprettur was able to persist output in spite of me restarting test-output-staging twice. Nice work.

busltee/transport.go
```diff
@@ -44,29 +50,28 @@ func (t *Transport) tries(req *http.Request) (*http.Response, error) {
 }
 
 type bodyReader struct {
-	r          io.Reader
-	w          *bytes.Buffer
-	bufferSent bool
+	r io.Reader
+
+	buffer        []byte
```
joshwlewis: Would this make sense as a temp file instead? These logs have the potential to be pretty large and could soak up quite a lot of memory. It might make sense given that in most cases, we won't even use this data due to network interruptions and restarts being fairly infrequent.

dmathieu: That'd make sense, yeah. We don't really care about the added latency due to reading from a file instead of memory. Good idea.

busltee/transport.go
```diff
@@ -4,69 +4,116 @@ import (
...
+	if b.buffReader == nil {
+		n, err := b.streamer.Read(p)
+		if err != nil {
+			return n, err
+		}
+		if rn, err := b.buffWriter.Write(p[:n]); err != nil {
+			return rn, err
 		}
 		return n, err
 	} else {
```
joshwlewis: Does the code in this else block even get excercised now? Looks like `b.buffReader` is always nil?

dmathieu: No. When we generate a new bodyReader, we'll read the content of `buffWriter` and put them into `buffReader`.
This value is nil if we've already read the buffer, or if there wasn't any.

busltee/transport.go
```diff
@@ -4,69 +4,116 @@ import (
...
-type bodyReader struct {
-	r          io.Reader
-	w          *bytes.Buffer
-	bufferSent bool
+func newBodyReader(streamer io.Reader, buffer *os.File) (*bodyReader, error) {
+	data, err := readBuffer(buffer)
+	if err != nil {
+		return nil, err
+	}
+	return &bodyReader{streamer, buffer, data}, nil
```
dmathieu: @joshwlewis buffReader is set here.

busltee/transport.go
```diff
@@ -16,40 +17,43 @@ type Transport struct {
...
 
 func (t *Transport) RoundTrip(req *http.Request) (res *http.Response, err error) {
-	tmpfile, err := ioutil.TempFile("", "busltee_buffer")
+	tmpFile, err := ioutil.TempFile("", "busltee_buffer")
 	if err != nil {
 		return nil, err
 	}
-	defer tmpfile.Close()
-	t.body = &bodyReader{req.Body, tmpfile, nil}
+	t.buffer = tmpFile
```
ojacobson: Does `Transport` guarantee that it'll close `Closer`-able `buffer`s?

busltee/transport.go
```diff
@@ -4,69 +4,116 @@ import (
...
+	Transport     http.RoundTripper
+	SleepDuration time.Duration
+
+	body   io.ReadCloser
+	buffer *os.File
 }
 
 func (t *Transport) RoundTrip(req *http.Request) (res *http.Response, err error) {
-	t.body = &bodyReader{req.Body, &bytes.Buffer{}, true}
+	tmpFile, err := ioutil.TempFile("", "busltee_buffer")
```
ojacobson: What drove the decision to buffer to disk?

joshwlewis:
Thanks for this. It looks good to me. 👍

busltee/transport.go
```diff
@@ -4,69 +4,116 @@ import (
...
+		return nil, err
+	}
+	return &bodyReader{streamer, buffer, data}, nil
 }
 
-func (*bodyReader) Close() error { return nil }
-func (b *bodyReader) Reset() {
-	b.bufferSent = false
+func readBuffer(b *os.File) (*bytes.Buffer, error) {
+	f, err := os.Open(b.Name())
```
ojacobson: is `f` ever closed?

busltee/transport.go
```diff
@@ -4,69 +4,116 @@ import (
...
-	req.ContentLength = -1
+	if t.SleepDuration == 0 {
+		t.SleepDuration = time.Second
+	}
 	return t.tries(req)
 }
 
 func (t *Transport) tries(req *http.Request) (*http.Response, error) {
-	res, err := t.Transport.RoundTrip(req)
+	req.Body.Close()
```
ojacobson: On the first call to `tries`, aren't `req.Body` and `t.body` pointers to the same object? Does this `Close()` close both, or does it somehow only close one? I ask, because the next line calls a function that reads from `t.body`…

busltee/transport.go
joshwlewis: Conversation here: https://github.com/heroku/busl/pull/97#discussion_r87329987

dmathieu: See Josh's comment earlier.
There can be a lot of buffered data, and storing that in memory is a path to issues.

busltee/transport.go
dmathieu: Yes, but the first request's body is `stdin`, with a `NopCloser` on top.
So the first close has no effect.

busltee/transport.go
dmathieu: It wasn't. Thanks.

busltee/transport.go
dmathieu: We can guarantee a close in the method's defer, as it won't be called until all tries have happened. Fixed that.

busltee/transport.go
ojacobson: Thanks.

busltee/transport.go
ojacobson: Ah. That's… somewhat less than totally obvious from the code, but it will work. Carry on.

busltee/transport.go
```diff
@@ -4,69 +4,118 @@ import (
...
 
 	if err != nil || res.StatusCode/100 != 2 {
 		if t.retries < t.MaxRetries {
+			time.Sleep(t.SleepDuration)
 			t.retries += 1
-			t.body.Reset()
-			return t.tries(req)
+			return t.tries(newReq)
 		} else {
 			return nil, err
```
joshwlewis: If you hit max retries on abnormal status codes, we'd be returning `nil, nil` here. In my testing, http.Client doesn't really appreciate that (pointer dereference). Can we return the last `res` here? In fact, you could just eliminate this whole else block with that approach.

PR #98, opened 2016-11-21
Upgrade Go

tt:


dmathieu:
LGTM

dmathieu:
As a side note though, I don't think deploying this to the kernel app is worth the effort. We can ship it after the migration to platform.

PR #99, opened 2016-11-30
Ping the publishers regularly

dmathieu:
The router requires sending data back to the client, or it will send an [H15](https://devcenter.heroku.com/articles/error-codes#h15-idle-connection), causing a disconnection every 55 seconds.

This sends an empty byte to the response every second. I have confirmed it does prevent the H15s.


tt:
Is this valid HTTP, @ferd? Can you really start sending a response before the request has been received?

It seems like a bug if the router disconnects a client that's actively writing a request body. Is this is Dogwood specific behavior?

tt:
Scrap that, this is the subscriber side not the publisher side.

tt:
@dmathieu, did you figure out why the other keepalive mechanism didn't work?

dmathieu:
The other keepalive mechanism does work. It's only for subscribers though.
This is for publishers.

dmathieu:
> Scrap that, this is the subscriber side not the publisher side.

This is actually the publisher side.

tt:
The title says "Ping the subscribers regularly".

tt:
I reinstate my previous comment. @ferd to the rescue.

ferd:
@tt you can send a response before the full request has been received (i.e. a 403 before a 150mb file upload has made it through), but you wouldn't send a response before having had a few bytes to figure out what the request is. Although nothing really prevents it, it would be a bit weird and I don't know that most clients would react well.

ferd:
Uh also the 55 seconds of inactivity should not trigger until the request body is sent fully, now that I re-read this. 

In fact the timeouts work like this:

- the router sends the request. Once the request is sent, then it starts a 30s timer to respond before an H12 is generated
- once bytes from the response have been received, you have 55s to send bytes here and there to prevent getting disconnected with an H15
- the 55s applies also to total line silence (i.e. during the request, you don't see data for 55s either uploaded for an H28 or downloaded for an H15)

dmathieu:
Does that mean this is going to be fixed in hermes?

ferd:
The behaviour I have described is how hermes works today.

tt:
@dmathieu, this might be a bug in the Dogwood router.

dmathieu:
This is specific to spaces. I thought this PR was fixing it, as I was testing it in the common runtime. But it doesn't when testing it in a space.

PR #100, opened 2016-12-14
Refactor keepalive into a generic package and send pings in busltee

dmathieu:
This refactors the KeepAlive Reader into it's own package so it can be used by busltee as well.

I've also had to refactor the way it works.
The old reader didn't support bytes smaller than the data it needs to hold.


tt:
This is destructive; it adds null bytes into the data as received by busl and that's what we'll ultimately persist. I think the current design where it will reconnect whenever disconnected by the router is actually better.

dmathieu:
Except that design doesn't work. The client gets a disconnection with an HTTP 200 (sent by the router), and doesn't know whether the feed has finished streaming, or if does need to reconnect.

PR #101, opened 2016-12-14
Godep => govendor

freeformz:


bin/go-version-sync-check.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 travis=$(cat .travis.yml | grep '^go:' | sed 's/^go: \([0-9]*\.[0-9]*\)\(\.[0-9]*\)\{0,1\}$/\1/')
-godep=$(cat Godeps/Godeps.json | grep 'GoVersion' | sed 's/.*"go\(.*\)".*$/\1/')
-diff <(echo $travis) <(echo $godep)
+goversion=$(< vendor/vendor.json jq -r '.heroku.goVersion' | sed 's/^go\(.*\)$/\1/')
+diff <(echo $travis) <(echo $goversion)
```
tt: There should be a end-of-file line break.

vendor/vendor.json
```diff
@@ -0,0 +1,60 @@
+{
+	"comment": "",
+	"ignore": "test",
```
tt: This should be sorted after "heroku". (It will happen automatically when the file changes but that introduces an unnecessary change in that pull request.)

vendor/vendor.json
```diff
@@ -0,0 +1,60 @@
...
+	"heroku": {
+		"install": ["./cmd/..."],
+		"goVersion": "go1.7"
+	},
+	"package": [
+		{
+			"checksumSHA1": "S+kZB+E4P+VhEzVhGTGZYWt0UkY=",
+			"comment": "0.4.0-11-g0cd5a1d",
```
tt: I wonder why govendor migrates these as comments. They should be `version` and `versionExact`.

freeformz: I think it only uses those when you fetch with a semver tag, like `govendor fetch github.com/heroku/rollbar@v0.3.0` produces:

```json
		{
			"checksumSHA1": "CD+EtSm6948GZ8+xmCRIZX8UcX0=",
			"path": "github.com/heroku/rollbar",
			"revision": "197e8a4fd165c96afe325b2297cf3bcaa5b9a6b4",
			"revisionTime": "2014-11-13T21:05:59Z",
			"version": "v0.3.0",
			"versionExact": "v0.3.0"
		},
```

PR #102, opened 2016-12-15
Run tests with -race

dmathieu:
I had to write this library: https://github.com/dmathieu/safe_buffer
Which introduces a very simple ReadWriter using a mutex.


busltee/transport_test.go
```diff
@@ -45,7 +47,7 @@ func TestDisconnection(t *testing.T) {
 
 	var callCount int
 	var expectedBody string
-	var stdin bytes.Buffer
+	stdin := safe_buffer.NewMock()
```
tt: According to [this blog post](https://blog.golang.org/package-names), "good package names are short and clear. They are lower case, with no under_scores or mixedCaps."

dmathieu: Would you have a name suggestion?

tt: Just "safebuffer"? Not that different from "httputil" and other packages.

dmathieu: Sounds good to me. I've renamed the package and rebased this branch.

PR #103, opened 2016-12-16
Add long streaming example

dmathieu:
I've been using this example to test longer streaming without any disconnection.


example/stream.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+APP_NAME=damien-busl
```
tt: This is local configuration. Can you maybe let the URL use `$BUSL_HOST` (to allow running it anywhere) and then put your value in .env (uncommitted) instead?

PR #104, opened 2016-12-16
Only close the writer if it didn't finish with an error

dmathieu:
When getting an unexpectedEOF, or any connection error, we shouldn't close the stream.

### DIFF
--- a/server/endpoints.go
+++ b/server/endpoints.go
@@ -41,7 +41,6 @@ func (s *Server) publish(w http.ResponseWriter, r *http.Request) {
 		handleError(w, r, err)
 		return
 	}
-	defer writer.Close()
 
 	body := bufio.NewReader(r.Body)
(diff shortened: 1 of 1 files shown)


tt:
Should there be a test so we don't reintroduce this?

dmathieu:
I agree. But sending a fake `unexpectedEOF` is quite tricky in local tests.

tt:
Can't you implement a type like this and assign it to `Body`:

    type errorReader struct {
    	err error
    }

    func (r *errorReader) Read(p []byte) (int, error) {
    	return 0, r.err
    }

dmathieu:
After discussing this, we're going without a test. We'd need to call `#publish` directly to be able to override the request body.
But we cannot do that, as we'd be missing context for `gorilla/mux`, namely the URI params and would be getting 404s as no stream id would be available.

PR #105, opened 2016-12-19
Block closing the reader until the current read has unblocked

dmathieu:
`Readers` are blocking, making it so that when we get a disconnection, the reader is still listening on stdin until it gets some data.
`#Close` isn't blocking though. So we think we closed the connection and can move on to the next one, when we're still reading for data in a connection which finished.

This mutex makes `#Close` blocking, so closing the body waits until the next bit has been received and written to the buffer.
The next connection can then kick in and not miss any data.

### DIFF
--- a/busltee/transport_test.go
+++ b/busltee/transport_test.go
@@ -2,6 +2,7 @@ package busltee
 
 import (
 	"bytes"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


tt:
What if the `Read` hangs? Isn't it an important property to be able to always close the connection?

What's the consequence of closing it without the synchronization? Shouldn't that itself end the `Read` call?

dmathieu:
The thing is, if `Read` hangs, it also blocks, and will wait until it receives a byte.
The effect is that we close the connection. But are still waiting on some data to read, which will be buffered
Then, as there's already something listening on stdin, the next connection doesn't get any data.

This mutex causes `Close` to block until `Read` has finished blocking, meaning we don't actually open a new connection and don't start listening on stdin.
Then the new connection only opens once the previous one isn't reading anymore, and properly gets all data from the buffer and the stream.

busltee/transport_test.go
```diff
@@ -128,3 +129,40 @@ func TestHTTPError(t *testing.T) {
 		t.Fatalf("was expecting 9 retries. Got %d", callCount)
 	}
 }
+
+type slowBuffer struct {
+}
```
tt: Can you move this to the previous line?

busltee/transport_test.go
```diff
@@ -128,3 +129,40 @@ func TestHTTPError(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer buffer.Close()
+
+	reader, err := newBodyReader(streamer, buffer)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	go func() {
```
tt: Maybe just `go reader.Close()`?

busltee/transport_test.go
dmathieu: :+1:

busltee/transport_test.go
dmathieu: Nice catch. Done.

PR #108, opened 2017-01-09
Separate the streamer and connect retries count

dmathieu:
We need to increase the streamer retries count, but not the connect count.


cmd/busltee/main.go
```diff
@@ -51,6 +51,7 @@ func parseFlags() (*cmdConfig, *busltee.Config, error) {
 	// Connection related flags
 	flag.BoolVarP(&publisherConf.Insecure, "insecure", "k", false, "allows insecure SSL connections")
 	flag.IntVar(&publisherConf.Retry, "retry", 5, "max retries for connect timeout errors")
+	flag.IntVar(&publisherConf.StreamRetry, "stream-retry", 60, "max retries for streamer disconnections")
```
dmathieu: I've set this to 60, as we can get disconnected once every minute if nothing gets logged. And builds last up to 60 minutes.

PR #113, opened 2017-01-24
Fail fast if redis returns an error

dmathieu:
Right now, if the `EXEC` fails, we will get a nil `list`, which causes panic errors.


broker/io.go
```diff
@@ -181,9 +181,21 @@ func (r *reader) fetch(length int) ([]byte, error) {
 	conn.Send("EXPIRE", r.channel.id(), redisChannelExpire)
```
tt: I think this can also return an error.

PR #114, opened 2017-01-24
Redis connections leak

dmathieu:
We cannot use `NopCloser` on readers, as it would nullify the `Close` method, and we would never release the redis connection.

### DIFF
--- a/broker/redis.go
+++ b/broker/redis.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"net/url"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/garyburd/redigo/redis"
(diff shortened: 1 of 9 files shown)


broker/redis.go
```diff
@@ -14,11 +15,41 @@ import (
...
-	redisPool          *redis.Pool
+	redisPool          *Pool
 	redisKeyExpire     = 60 // redis uses seconds for EXPIRE
 	redisChannelExpire = redisKeyExpire * 60
 )
 
+type Pool struct {
+	*redis.Pool
+
+	m *sync.Mutex
```
tt: I would suggest `mu`. I have at least seen that name in a number of codebases so it's more obvious what it is.

dmathieu: Done.

broker/redis.go
```diff
@@ -14,11 +15,41 @@ import (
...
+	c int
+}
+
+func (p *Pool) Get() Conn {
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	p.c += 1
+	util.CountWithData("redis.conn.get", 1, "conn_count=%d, caller=%q#%d", p.c)
```
tt: Rather than `conn_count`, can you emit a `sample#...connections` so we can track it in Librato?

broker/redis.go
```diff
@@ -14,11 +15,41 @@ import (
 var (
 	redisURL           = flag.String("redisUrl", os.Getenv("REDIS_URL"), "URL of the redis server")
 	redisServer        *url.URL
-	redisPool          *redis.Pool
+	redisPool          *Pool
 	redisKeyExpire     = 60 // redis uses seconds for EXPIRE
 	redisChannelExpire = redisKeyExpire * 60
 )
 
+type Pool struct {
```
dmathieu: We can, but then other packages will use a private struct to acquire connections.
Unfortunately, redigo doesn't provide a Pool interface.

broker/redis.go
dmathieu: Done.

broker/redis.go
dmathieu: You're right, it isn't. I've made this struct private.

PR #115, opened 2017-01-25
Only keep the mutex lock for the duration of setting the variable

dmathieu:
Acquiring the connection can take some time, which causes timeouts.

### DIFF
--- a/broker/redis.go
+++ b/broker/redis.go
@@ -5,7 +5,7 @@ import (
 	"log"
 	"net/url"
 	"os"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/garyburd/redigo/redis"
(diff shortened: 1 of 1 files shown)


tt:
How about using [`atomic.AddInt64`](https://golang.org/pkg/sync/atomic/#AddInt64) instead?

broker/redis.go
```diff
@@ -28,12 +28,23 @@ type pool struct {
 }
 
 func (p *pool) Get() Conn {
+	n := p.incCounter()
```
tt: I think this would be `n := atomic.AddInt64(&p.c, 1)` using that package.

tt:
This obviously makes sense to do but why would we wait a long time to get a connection? Is the pool not configured appropriately for busl?

broker/redis.go
dmathieu: Nice, I didn't know about that. Changed.

dmathieu:
The pool opens connections as needed, apparently starting with a single one.
So it would make sense that it starts by opening a lot of them, which definitely causes issues if it can't happen concurrently.

PR #121, opened 2017-01-30
Use well-known configuration variable

tt:
#109 was half right but also half wrong.


tt:
I set the configuration on busl and busl-staging already.

dmathieu:
LGTM

PR #122, opened 2017-02-01
Vary storage by host

tt:
If we set the descriptive `TEST_OUTPUT_HEROKU_COM_STORAGE_BASE_URL` configuration variable, this will allow @heroku/devex to retire their instance.


tt:
@dmathieu, done!

dmathieu:
LGTM

joshwlewis:
For us to use this, we'd need to

1) Grant your busl user access to our bucket
2) Set the TEST_OUTPUT_HEROKU_COM_STORAGE_URL
2) Configure DNS/SSL for test-output.heroku.com to point to busl
3) Switch our services to point use test-output.heroku.com

Does that sound about right?

dmathieu:
@joshwlewis almost, yes. Busl has no IAM user. You need to pass presigned credentials to busl when publishing and creating a stream.

Vacuole does that [here](https://github.com/heroku/vacuole/blob/f19c3d6ea0ba46629b0aad9f77ebdde2ff38145c/lib/vacuole/data.rb#L135).

gudmundur:
@tt any news on this? We'd love to move over.

PR #124, opened 2017-02-09
Stop reading the body once it has been closed

dmathieu:
We still looped through all entries in a body until that streamer finished all data.
Once a body has been closed, it means it's connection isn't open anymore, and we can stop reading.

Among a lot of disconnections, this could have caused latencies, as there were a lot of bodies open, all of which acquiring a lock on the mutex to know if they could close themselves or not.

### DIFF
--- a/busltee/transport.go
+++ b/busltee/transport.go
@@ -104,7 +104,7 @@ func (t *Transport) newBodyReader() (io.ReadCloser, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &bodyReader{reader, t}, nil
+	return &bodyReader{reader, t, false}, nil
 }
 
 func (t *Transport) Close() error {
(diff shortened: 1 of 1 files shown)


busltee/transport.go
```diff
@@ -139,13 +139,19 @@ func (w *broadcastingWriter) Write(p []byte) (int, error) {
 
 type bodyReader struct {
 	io.ReadCloser
-	t *Transport
+	t      *Transport
+	closed bool
+}
+
+func (b *bodyReader) Close() error {
+	b.closed = true
```
ojacobson: I'm nervous about setting the flag before determining whether the underlying close operation has succeeded or not. I realize that failing to close a read-only handle is pretty rare, so it's _probably_ academic, but what does the underlying transport's `IsClosed()` method return if `Close()` returns an error?

dmathieu: Nice catch. I've fixed it.

PR #125, opened 2017-02-09
Signal readers

tt:


busltee/transport.go
```diff
@@ -109,28 +111,48 @@ func (t *Transport) Close() error {
...
 	return nil
 }
+
 func (t *Transport) isClosed() bool {
 	t.mutex.Lock()
 	defer t.mutex.Unlock()
 	return t.closed
 }
 
+type nopLocker struct{}
```
ojacobson: - [ ] Should this be used somewhere?
- [ ] Should this have a type assertion?

ojacobson:
Basically looks good. I have a couple of Q's (see inline) but I'm happy with the structure.

busltee/transport.go
tt: It's used on line 34: `sync.NewCond(&nopLocker{})`

The [`sync.Cond`](https://golang.org/pkg/sync/#Cond) type requires a lock but we don't require it as each reader reads from the file.

ojacobson: 👍  thanks!

PR #126, opened 2017-02-14
Buffer the sse content in a byte array smaller than the one provided

dmathieu:
SSE adds data to the bytes array. Therefore, if the data we add goes over `len(p)`, some will be missed, and we won't serve it.
That causes invalid streams, which never get rendered by the dashboard.

### DIFF
--- a/encoders/sse.go
+++ b/encoders/sse.go
@@ -46,6 +46,10 @@ func (r *sseEncoder) Read(p []byte) (n int, err error) {
 
 	if n > 0 {
 		buf := format(r.offset, q[:n])
+		if len(buf) > len(p) {
+			return 0, errors.New("buffer length cannot be higher than bytes array")
+		}
+
 		r.offset += int64(n)
 		n = copy(p, buf)
 	}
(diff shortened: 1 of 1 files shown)


tt:
Should we return an error if it ever happens that we want to write more than we can?

encoders/sse.go
```diff
@@ -39,14 +39,13 @@ func (r *sseEncoder) Seek(offset int64, whence int) (n int64, err error) {
...
-// length of data to be read.
 func (r *sseEncoder) Read(p []byte) (n int, err error) {
-	n, err = r.ReadCloser.Read(p)
+	// We assume SSE won't add more than twice the amount of data we get
+	q := make([]byte, len(p)/2)
+	n, err = r.ReadCloser.Read(q)
 
 	if n > 0 {
-		buf := format(r.offset, p[:n])
+		buf := format(r.offset, q[:n])
```
tt: I think `if len(buf) > len(r) { return 0, <error> }` would make sense to prevent us from not noticing again.

dmathieu: Good idea. I've added it.

tt: :+1:

PR #128, opened 2017-03-15
Add a request id to all responses

dmathieu:
Take the request id provided with the request if one is there. Otherwise, generate a new uuid.


server/middleware.go
```diff
@@ -62,6 +62,13 @@ func (s *Server) addDefaultHeaders(fn http.HandlerFunc) http.HandlerFunc {
 		if origin := r.Header.Get("Origin"); origin != "" {
 			w.Header().Set("Access-Control-Allow-Origin", origin)
 		}
+
+		requestID := r.Header.Get("Request-ID")
```
tt: Should we also use this for the logging context?

dmathieu: We already get the request id when logging it.
https://github.com/heroku/busl/blob/master/util/response_logger.go#L38

PR #130, opened 2017-03-20
Log request id and key to rollbar

dmathieu:
The current information we get from rollbar occurences doesn't allow us to properly investigate the behavior on builds.

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -7,6 +7,7 @@ import (
 	"github.com/heroku/busl/broker"
 	"github.com/heroku/busl/storage"
 	"github.com/heroku/busl/util"
+	"github.com/heroku/rollbar"
 )
 
 var errNoContent = errors.New("No Content")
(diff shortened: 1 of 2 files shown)


server/errors.go
```diff
@@ -52,3 +53,10 @@ func handleError(w http.ResponseWriter, r *http.Request, err error) {
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+func logError(req *http.Request, err error) {
+	rollbar.ErrorWithExtras(rollbar.ERR, err, map[string]interface{}{
+		"request_id": req.Header.Get("Request-Id"),
+		"build_id":   key(req),
```
tt: Should this maybe be just `key`? Remember busl will serve more things than builds.

dmathieu: You're absolutely right, thank you.

PR #133, opened 2017-04-11
Handle publishers sending no data at all

dmathieu:
We need to handle this to be able to close the stream without adding anything to it, which errors out as we get an EOF in the discard.

See https://rollbar.com/Heroku-3/busl/items/39/


server/endpoints.go
```diff
@@ -51,7 +51,7 @@ func (s *Server) publish(w http.ResponseWriter, r *http.Request) {
 	}
 	if wl > 0 {
```
tt: Why is `wl` not zero if the stream is empty?

dmathieu: Because `wl` is the length of the data already sent.
We are closing the stream without sending any new data. But some data has already been sent.

tt: Ah, so it's not because we're not sending any new data but because we also fail to provide the same data as busl would otherwise expect. I think we need to use a different endpoint then. How about adding `PUT` that doesn't have the same expectation? Then we can still fail properly if any subscriber forgets to resend the same data.

dmathieu: We could also add `DELETE` to just close the stream.

dmathieu: We can't use the `PUT` verb. It's already used to create a stream.

dmathieu:
Closing. #134 should be enough for our current need.

PR #134, opened 2017-04-11
Add DELETE endpoint to close a stream

dmathieu:
This endpoint allows us to close a stream without sending any new data to it.

### DIFF
--- a/server/endpoints.go
+++ b/server/endpoints.go
@@ -111,3 +111,20 @@ func (s *Server) subscribe(w http.ResponseWriter, r *http.Request) {
 	}
 	util.CountWithData("server.sub.read.finish", 1, "msg=%q request_id=%q", err, r.Header.Get("Request-Id"))
 }
+
+func (s *Server) closeStream(w http.ResponseWriter, r *http.Request) {
+	writer, err := broker.NewWriter(key(r))
+	if err != nil {
+		handleError(w, r, err)
+		return
+	}
+
+	util.CountWithData("server.close", 1, "request_id=%q", r.Header.Get("Request-Id"))
+	err = writer.Close()
+	if err != nil {
+		handleError(w, r, err)
+		return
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -58,6 +58,7 @@ func (s *Server) router() http.Handler {
 	r.HandleFunc("/streams/{key:.+}", s.addDefaultHeaders(s.subscribe)).Methods("GET")
 	r.HandleFunc("/streams/{key:.+}", s.addDefaultHeaders(s.publish)).Methods("POST")
 	r.HandleFunc("/streams/{key:.+}", s.auth(s.addDefaultHeaders(s.createStream))).Methods("PUT")
+	r.HandleFunc("/streams/{key:.+}", s.auth(s.addDefaultHeaders(s.closeStream))).Methods("DELETE")
```
tt: Why do we need to `auth` this?

server/endpoints.go
```diff
@@ -111,3 +111,20 @@ func (s *Server) subscribe(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	util.CountWithData("server.close", 1, "request_id=%q", r.Header.Get("Request-Id"))
+	err = writer.Close()
+	if err != nil {
+		handleError(w, r, err)
+		return
+	}
+	// Asynchronously upload the output to our defined storage backend.
+	go storeOutput(key(r), requestURI(r), s.StorageBaseURL(r))
```
tt: Do we need to do this asynchronously?

server/server.go
dmathieu: You're right, we don't.

server/endpoints.go
dmathieu: Not sure we necessarily do. But we do it asynchronously in `publish`. I guess to speed up the response time.

PR #136, opened 2017-04-19
Don't require a chunked transfer encoding to publish data

dmathieu:
The request body will always be a reader.
So our interface will always work, whether there is a chunked transfer encoding or not.

While streaming data will always need chunked requests, we're starting to close streams with error messages coming from other components.
Those are static errors which don't make sense as chunked requests.


dmathieu:
@ojacobson this should reduce the complexity in heroku/vacuole/pull/482

PR #138, opened 2017-05-11
Kill the entire child process group once the child exits.

ojacobson:
This prevents the child from starting orphans that then hold stdout or stderr open indefinitely. Since we're nominally managing the child, exiting when the child exits is reasonably sensible.

We may cut off part of the child process' output when this happens.


dmathieu:
See #139

PR #147, opened 2017-06-27
Upgrade Rollbar client lib.

ojacobson:
This actually broke something for me - I attempted to set up busltee in a docker image and it caught the updated rollbar lib from my GOPATH, breaking the build. Might as well update and fix everything…


tt:
I'm always fine upgrading but I wonder how it actually broke; why did it not use the vendored version when you built it?

PR #152, opened 2017-07-25
Ignore `SIGPIPE`

tt:
I decided to ignore `SIGPIPE` because it's a signal sent to the process reading/writing from a broken pipe. Therefore it doesn't make sense to forward it to a child process.

### DIFF
--- a/busltee/runner.go
+++ b/busltee/runner.go
@@ -226,8 +226,17 @@ func deliverSignals(cmd *exec.Cmd) {
 	signal.Notify(sigc)
 	go func() {
 		s := <-sigc
-		logrus.WithFields(logrus.Fields{"busltee.signal.deliver": s}).Info("OK")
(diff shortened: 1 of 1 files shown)


tt:
I had initially included two other changes in this:

1. Stopping the `deliverSignals` goroutine when the process exists.
2. Iterating over the received signals so multiple are delivered.

It might be worth tackling these later.

busltee/runner.go
```diff
@@ -226,8 +226,17 @@ func deliverSignals(cmd *exec.Cmd) {
 	signal.Notify(sigc)
 	go func() {
 		s := <-sigc
```
dmathieu: You will need a for loop, as we can now get several signals.

tt: We could before as well. I had included it but that's really a separate change that bears more risk.

PR #158, opened 2017-09-08
Allow logging arbitrary fields

dmathieu:
This brings #144 in a more structured way.


busltee/buslteelogger/main.go
```diff
@@ -0,0 +1,65 @@
...
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// LogFields stores all the custom fields
+type LogFields []string
```
dmathieu: This would probably better as a `map[string]string`. But somehow, I'm unable to append data to it in `Set` with that type.

busltee/buslteelogger/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (l *LogFields) String() string {
+	return fmt.Sprintf("%q", *l)
+}
+
+// Set is used by the flag package to set new values
+func (l *LogFields) Set(value string) error {
+	s := strings.Split(value, "=")
```
ojacobson: `SpitN`, to allow fields with = in the value, perhaps?

busltee/runner.go
```diff
@@ -39,14 +40,14 @@ func Run(url string, args []string, conf *Config) (exitCode int) {
 	done := post(url, reader, conf)
 
 	if err := run(args, writer, writer); err != nil {
-		logrus.WithFields(logrus.Fields{"count#busltee.exec.error": 1}).Error(err)
+		buslteelogger.WithFields(logrus.Fields{"count#busltee.exec.error": 1}).Error(err)
```
ojacobson: This feels like a really long package name for something used as widely as logging.

dmathieu: We could bring it to a `logger` package, though this is pretty busltee specific.
I've been doing an external package, with the idea in mind that busltee is currently very CLI-command specific, and as we start needing to use it as a Go package (providing it a buffer instead of stdin for example), we will probably need to split into a few smaller packages.

I've also named this package `buslteelogger` due to the subpackages convention such as [httptest](https://golang.org/pkg/net/http/httptest/).

tt: If it's busltee specific, why does it need a separate package?

busltee/buslteelogger/main.go
tt: Can you try again? It should be possible. I think you might need parentheses, though.

busltee/runner.go
dmathieu: It doesn't need per-se. Having it in a separate package allows us to namespace the logs instead of calling directly `Errorf`.

tt: It also means that other people could import that package and use it. I'm in favor of minimizing the public interface.

busltee/buslteelogger/config.go
```diff
@@ -0,0 +1,64 @@
...
+
+// ConfigureLogs configures the log file
+func ConfigureLogs(logFile string, fields LogFields) {
+	lf := logrus.Fields{}
+	for k, v := range fields {
+		lf[k] = v
+	}
+
+	l = &logger{output(logFile), lf}
+	//logrus.SetOutput(l.out)
```
tt: Is this a mistake?

dmathieu: Yes it is. I've removed it.

busltee/runner.go
dmathieu: I've moved buslteelogger to the busltee package.

PR #159, opened 2017-09-28
Write header on error

tt:
We correctly return to prevent closing the stream but it turns out the `status=200` could be a red herring.


server/endpoints.go
```diff
@@ -56,6 +56,7 @@ func (s *Server) publish(w http.ResponseWriter, r *http.Request) {
 
 	if err == io.ErrUnexpectedEOF {
 		util.CountWithData("server.pub.read.eoferror", 1, "msg=%q request_id=%q", err, r.Header.Get("Request-Id"))
+		w.WriteHeader(http.StatusBadRequest)
```
dmathieu: Should we use `handleError` instead to also return a body?

tt: I honestly think it's clearer not to. It would require a new `case` to cover it and that's just an indirection; now I need to read two things to know how the endpoint behaves. There will also not be anyone to see the body as the client disconnecting is the cause of the issue.

PR #160, opened 2017-10-17
Scrub URLs

tt:


busltee/logger.go
```diff
@@ -91,3 +92,23 @@ func logFields() *logrus.Entry {
...
 	return logrus.WithFields(l.defaultFields)
 }
+
+type scrubber struct {
+	logrus.Formatter
+}
+
+func (s *scrubber) Format(entry *logrus.Entry) ([]byte, error) {
+	for key, value := range entry.Data {
+		if key == "url" {
```
dmathieu: Could we use a regex on the value to detect something that looks like an URL instead? That would prevent fields logged as something different than url from having creds.

tt: I felt it was too much complexity for this package but I changed it regardless. I guess we can extract it to a general package later.

PR #161, opened 2017-10-23
Increase time-out

tt:
This increases the time we try publishing a stream before timing out.

@benburkert discovered that the current time-out is the same time [we back off between stream retries](https://github.com/heroku/busl/blob/1b7fbe9ebc8de3eb74ca1a6e3f9a101f74a06191/busltee/transport.go#L63). That means an unfortunately timed error might prevent the stream closing properly.


dmathieu:
@tt should we ship this?

tt:
Yes!

PR #164, opened 2017-11-08
Always send a 204 if the stream is empty

dmathieu:


dmathieu:
@tevanoff @heroku/build I've figured it out. `NoContent` was only checking on redis, but when the stream is archived, it's not there anymore and we'd always ignore the method.

I've updated the behavior to look for how many bytes were sent in `io.Copy`, and send a 204 if that was 0.

dmathieu:
[io.Copy](https://golang.org/pkg/io/#Copy)'s first argument is the number of bytes written. So that assumption seems safe to assume 😜

server/endpoints.go
```diff
@@ -105,6 +105,11 @@ func (s *Server) subscribe(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		logError(r, err)
 	}
+	if n == 0 {
+		handleError(w, r, errNoContent)
```
tt: Treating this as an error is a bit odd.

dmathieu: Indeed. I've fixed that in #165.

PR #166, opened 2017-11-16
Implement stream buffer struct

dmathieu:
This buffer allows keeping a streamer open until we close it.
This way, we can send data to busltee without relying on stdin, and keep sending it until we decide to close.

This has been reused in a couple internal projects. Extracting it here allows easier sharing between them.


busltee/buffer_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"io/ioutil"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStreamBuffer(t *testing.T) {
+	t.Run("has not received any data yet", func(t *testing.T) {
```
tt: I think these should be different tests when they're not sharing any context although I agree it's nice being able to use strings for descriptions.

dmathieu: Sure. I've done that.

PR #170, opened 2018-01-12
busltee: add 30s pinging to http2 uploads

benburkert:
Configure the Transport type to automatically send http2 ping frames when the underlying connection is over h2. The pings will prevent middle-boxes from closing idle connections when an uploads stalls due to slow input to stdin.


cyx:
This looks very neat!

PR #172, opened 2018-01-28
Fix Makefile typo

WillAbides:
The sample .env file is named .sample.env, not .env.sample


tt:
@WillAbides, thank you for this contribution.

Would you mind renaming the file instead of changing the Makefile? The filename ".env.sample" is what we generally use so I would prefer uniformity.

dmathieu:
Thank you @WillAbides! 🌷

PR #177, opened 2018-04-18
Filter request-ids

dmathieu:
So consumers can't inject unwanted data as Request-IDs.

cc @brompwnie


server/middleware.go
```diff
@@ -64,7 +69,7 @@ func (s *Server) addDefaultHeaders(fn http.HandlerFunc) http.HandlerFunc {
 		}
 
 		requestID := r.Header.Get("Request-ID")
-		if requestID == "" {
+		if requestID == "" || !uuidPattern.MatchString(requestID) {
```
tt: A request ID is not required to be a UUID:

> The value must be between 20 and 200 characters, and consist of ASCII letters, digits, or the characters `+`, `/`, `=`, and `-`.

(See [the "how it works" section of the Dev Center article on request IDs](https://devcenter.heroku.com/articles/http-request-id#how-it-works).)

The header value might also be comma separated to include multiple request IDs.

dmathieu: I've updated the regex to allow all alphanumeric characters, as well as the special ones and to be 20 to 200 chars.

server/middleware.go
```diff
@@ -17,6 +18,10 @@ import (
 	"github.com/heroku/busl/storage"
 )
 
+var (
+	uuidPattern = regexp.MustCompile("^[A-Za-z0-9,+/=-]{20,200}$")
```
tt: This is no longer a UUID pattern.

tt:
You should make sure that we parse this value once so the same sanitization also applies here:

https://github.com/heroku/busl/blob/7f76e4e72cd95b7165f5a50853e278865db12a3c/internal/util/response_logger.go#L39-L41

dmathieu:
Good catch. I've done so.

internal/util/request_id.go
```diff
@@ -0,0 +1,26 @@
...
+	"regexp"
+)
+
+var (
+	requestIDPattern = regexp.MustCompile("^[A-Za-z0-9,+/=-]{20,200}$")
+)
+
+// GetRequestID retrieved the request id, or generates one
+func GetRequestID(req *http.Request) string {
+
```
tt: Remove this line break.

internal/util/request_id.go
```diff
@@ -0,0 +1,26 @@
...
+	id := req.Header.Get("Request-ID")
```
tt: Can you add the fallback to the `X-Request-ID` header?

dmathieu: Sure. Done.

PR #190, opened 2019-01-04
Bump HTTP timeout

tbalthazar:
Heroku CI tests are allowed to run for 2 hours.
Some customer with slow tests cannot see the output of their tests
because Busl seem to timeout and close the connection without persisting
the data.

Relate customer issue:
https://support.heroku.com/tickets/660731

Error in Rollbar:
https://rollbar.com/Heroku-3/busl/items/130/occurrences/62314961218/

Example error we see in the logs:

```
2018-12-21T20:07:35.712677+00:00 18.206.40.16 local7.info
heroku[router]: - e68e7ed9-5fb7-4d74-8085-41aa2f519e2d at=error code=H13
desc="Connection closed without response" method=POST
path="/streams/f2ba/test-runs/f2ba6e16-8def-4630-a2a6-516f949785b8"
host=test-output.heroku.com
request_id=ae88abe3-b297-381f-5bfd-ba3b1bb57670 fwd="54.197.172.106"
dyno=web.2 connect=0ms service=3600000ms status=503 bytes=561
protocol=http/1.1 tls_version=tls1.2
```

We already bumped the Redis channel timeout, but it didn't solve the
issue:
https://github.com/heroku/busl/pull/189


tbalthazar:
@dmathieu Thanks for chiming in. Do you think it's too risky to deploy this change to see if it fixes/mitigate the customer issue? We could then consider making the bigger change after it has been validated? wdyt?

PR #196, opened 2019-08-20
move to go mod

beanieboi:
move to go mod


Makefile
```diff
@@ -16,8 +16,9 @@ precommit: tidy test
 tidy: goimports
 	./bin/go-version-sync-check.sh
 	test -z "$$(goimports -l -d $(GO_FILES) | tee /dev/stderr)"
-	go vet ./...
-	dep status
+	go vet -mod=vendor ./...
+	go mod tidy
+	go mod verify
```
edmorley: This matches the approach Longshore uses, however I found with Longshore that it doesn't clean up the files from removed dependencies, and an additional `go mod vendor` should be added too:
https://github.com/heroku/longshore/pull/537#issuecomment-501256917

It seems like we should have a documented best-practice `tidy` step that all projects can use, rather than having to try and figure it out each time? @freeformz is this something you might be interested in doing? :-)

tt: I like this idea. It could be a package that you simply `go run github.com/heroku/go-tidy`.

beanieboi:
> Thank you for doing this - I look forward to when all of Lifecycle's components are using Go modules and I don't have to fight with GOPATH nonsense locally!

same same :D

PR #205, opened 2019-09-30
Replace requestId with requestID

dmathieu:
In order to make gofmt happy


dmathieu:
I run it locally with syntastic, yes. We can add it to travis, but there are currently so many failures that we'd be hours of work to fix them.

bensymonds:
Ah ok, I thought this was the only one 😂

dmathieu:
It's the only one in that package 😉

PR #209, opened 2019-10-01
Limit the output size

dmathieu:
This adds a new proxy writer, which monitors how many bytes have been sent to the output, and returns an error if there have been too many.
Then, the endpoint will add an error and close the stream.

I have set the limit to 300MB for now based on the maximums seen in the past 24h. We will probably need to wait for a few more days before shipping this though.


dmathieu:
> with this change, I'm presuming builds that hit the new limit will fail after the final "Output size limit reached" message?

Yes. The same is true for CI runs. Release phase dynos should keep running, but the output will only be available through `heroku logs`.

> after this lands should we document the maximum output size on https://devcenter.heroku.com/articles/limits ?

Yes we should.

> is the 300MB limit sufficient to stop the busl Redis memory exhaustion, or do you think we will need/want to lower it further after the initial rollout?

Increasing our memory right now is useless, as there's no limits. But once this is in place, doing it will make more sense, if that's required.
With the 8GB we have right now, we could have a bit over 20 outputs reaching the limit before having an incident ourselves.

> are there any other services using busl apart from codon/endosome? what about post-deploy output etc? (though I think nyata stores that in S3 and returns through its API iirc?)

Getting the output is publicly documented. So technically, anyone could do it. It shouldn't be a problem for any of them though. They'll just see the stream finish.
Profa and release-dyno-watcher are pushing to busl, the same way as codon. There isn't anything public doing it though.

beanieboi:
> > with this change, I'm presuming builds that hit the new limit will fail after the final "Output size limit reached" message?
> 
> Yes. The same is true for CI runs. Release phase dynos should keep running, but the output will only be available through `heroku logs`.

Could we display the message (maybe adding some ansi colors) and keep the build running so it would still pass even tho it produces too much log output. We would discard the log output after it reached the limit tho.

dmathieu:
We surely could. Though I would find the behavior of seeing the build output stop, but said build somehow releasing (or erroring with no information) later on to be quite weird.

bensymonds:
We also discussed storing the _tail_ of the output (rather than the _head_, which it sounds like this is doing). Then we could put an "output truncated"-type message at the beginning. Did you consider that? That way the user would (a) see it continue to scroll (b) see any errors etc which come at the end of the log.

Do we ever get input from product people on changes like this?

beanieboi:
> We also discussed storing the _tail_ of the output (rather than the _head_, which it sounds like this is doing). Then we could put an "output truncated"-type message at the beginning. Did you consider that? That way the user would (a) see it continue to scroll (b) see any errors etc which come at the end of the log.

I think this highly depends on your test suite. I think ruby puts them at the end, Golang puts them where they happen. I think either way is fine.

> Do we ever get input from product people on changes like this?
Yes, i believe @flxfxp is the current Build PM.

> Though I would find the behavior of seeing the build output stop, but said build somehow releasing (or erroring with no information) later on to be quite weird.

i'm not sure how many people watch builds. Something we observed at my previous job (Codeship a CI/CD provider) we found out that people mostly interact with the system when something fails.
e.g. a Email about a failed build was the main entry point into the build logs

If a build fails, i would not expect that would be because of log output truncation of the platform.
It would be great to find a middle ground. This means if we detect truncated logs, we notify the build author via Dashboard Notifcations and warn them about it, but don't fail the build.

server/endpoints.go
```diff
@@ -52,12 +55,14 @@ func (s *Server) publish(w http.ResponseWriter, r *http.Request) {
...
 
+	lw := broker.NewLimitWriter(writer, maxOutputSize, int(wl))
+
 	requestID := r.Header.Get("Request-Id")
 	util.CountWithData("server.pub.existing.bytes", wl, logrus.Fields{
 		"request_id": requestID,
 		"path":       r.URL.Path,
 	})
-	b, err := io.Copy(writer, body)
+	b, err := io.Copy(lw, body)
```
freeformz: FWIW: This can be done with an [io.LimitReader](https://golang.org/pkg/io/#LimitedReader) here instead of adding the custom limit writer.

freeformz: At least AFAICT.

dmathieu: It technically would, yes. But LimitReader returns `io.EOF` once the limit has been reached. So we wouldn't be able to know the limit had been reached and display the warning.

edmorley: It seems this might be worth adding as a comment?

dmathieu: The comment is in the writer itself: https://github.com/heroku/busl/pull/209/files#diff-08a030584b2a4c6dfbd8bf65e3b8725aR28-R29

edmorley: I mean as a clearer "this is why we're not using the built-in X"

bensymonds: Or maybe in the commit message? Just thinking if you made a code comment for everything like this you'd end up with a lot of comments...

edmorley: @bensymonds agreed, but I don't think it's uncommon to add a brief description of a new class/module/... (eg for Python in the class docstring) as to why that class/... exists in cases where it's not obvious.

A code comment is fine for now :-)

edmorley: > A code comment is fine for now :-)

Sorry I meant commit

bensymonds: Ok yeah also sorry from me - I thought you meant a code comment on this line (where it's called). I'm fine with it being mentioned at the top of the class/module/package definition (with usual caveat that it'll probably go out of date at some point..).

dmathieu: I've defined it as a godoc comment for the struct.

freeformz: @dmathieu 
```go
if err == io.EOF && n >= maxBytes {
  // display warning / notice
}
```

freeformz: or if you have a reference to the io.LimitReader...

``go
if err == io.EOF && lr.N <= 0 {
  // display warning / notice
}
```

Comment (freeformz) [reply server/endpoints.go]: FWIW: It doesn't make sense to me to have to maintain the LimitWriter code when something in the stdlib work sufficiently.

**Comment by ojacobson** [comment]
Something to consider: any limit we set _will_ be exceeded eventually (if we're successful as a business) and, in the long run, will be exceeded regularly (if we're successful for long enough). With a 300mb per-stream limit, it'd take around 24 concurrent maximum-size streams to take out Busl again. That's a lot more than we have today (averaging zero, with the occasional spike), but not a huge number in the grand scheme of things.

We can obviously move that number by using a larger Redis plan, but there's a limit to vertical scale in that sense. Similarly, we can move the number by running multiple independent Busl services, and that gets us more headroom, but there's nothing in place to actually do that today.

Is there anything we can do in Busl that would obviate this problem entirely, either by limiting based on total data stored or by eliminating the Redis step somehow?

**Comment by apghero** [comment]
I don't mean to helicopter in, but I'm curious if anyone has considered a plan that doesn't involve killing the build when the limit is reached, and instead using something like a circular buffer for the output that keeps only the last `$N` MB? This _seems_ like a more humane solution.

**Comment by bensymonds** [comment]
@apghero Yeah that's what I said higher up in https://github.com/heroku/busl/pull/209#issuecomment-536991504. I think doing that is less confusing, and will result in the more useful stuff (at the end of the log) being recorded. But that's just my opinion.

Or.... we keep the head and tail of the log and truncate the _middle_? That might be more difficult to implement in the current architecture though.

File: server/endpoints.go [hunk 2]
```diff
@@ -6,12 +6,15 @@ import (
...
 	"net"
 	"net/http"
+	"strings"
 
 	"github.com/heroku/busl/broker"
 	"github.com/heroku/busl/internal/util"
 	"github.com/sirupsen/logrus"
 )
 
+var maxOutputSize = 300 * 1024 * 1024
```
joshwlewis: Could we allow this to be configured via an ENV var? This seems like a reasonable default, but it might be helpful to be able to change it on the fly if/when we see a change in usage patterns.

dmathieu: I'm not sure that's a good idea. We will need to document the limit in https://devcenter.heroku.com/articles/limits, and making it changeable on the fly makes the documentation irrevelant quite easily.

broker/limit.go
```diff
@@ -9,7 +9,9 @@ import (
 var ErrLimitReached = errors.New("writer limit reached")
 
 // LimitWriter is a writer which allows a set maximum of bytes to be written,
-// and rejects writes afterwards
+// and rejects writes with `ErrLimitReached` afterwards.  We use a custom
+// struct and not `io.LimitReader` as we want to know when the limit has been
+// reached (as opposed to getting an `io.EOF`).
```
freeformz: I don't understand this. io.LimitReader returns an io.EOF when the limit is exhausted or the actual EOF is reached. AFAICT this is exactly the semantics we want. i.e. We want to process data until a limit is reached or the natural EOF is reached.

dmathieu: If we get an io.EOF, we can't know that we reached the limit and we can't add the error to the output.
We also can't return an HTTP error to the publisher, if it wishes to stop processing.

freeformzSFDC: yes, you do.

```go
lr := io.LimitReader{R: body, N: maxOutputSize}
b,err := io.Copy(writer, lr)
if err == io.EOF && lr.N <= 0 {
  // display warning / notice
  return
}
if err != nil {
  // return http error
}
```

dmathieu:
I've opened #211 as an alternative solution.

dmathieu:
#211 is replacing this PR.

PR #210, opened 2019-10-25
Upgrade Go to 1.13

edmorley:
https://golang.org/doc/go1.13

The vendor file removals are the result of `make tidy`.

Fixes [W-6833321](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007c0uNIAQ/view).


edmorley:
Travis is failing with errors like:

```

```

This appears to be due to  [CL 173722](https://go-review.googlesource.com/c/go/+/173722/), which according to the Go 1.13 [release notes](https://golang.org/doc/go1.13) means:

> Testing flags are now registered in the new Init function, which is invoked by the generated main function for the test. As a result, testing flags are now only registered when running a test binary, and packages that call flag.Parse during package initialization may cause tests to fail. 

And busl calls `flag.Parse()` during `init()` here:
https://github.com/heroku/busl/blob/b8a2c337ce97d86b27158512ffd8d18d1a36e9ba/broker/redis.go#L45-L52

The purpose of `init()` (something I wasn't familiar with) is described here:
https://golang.org/doc/effective_go.html#init

So sounds like we need to move the flag parsing (and thus also Redis connection pool setup) out of init and into TBD.

dmathieu:
Flag parsing shouldn't happen in a package. It should happen in the main one, and the value should be passed as configuration to the appropriate methods.

edmorley:
Agreed; though this now means quite a refactor of broker usage is required.

beanieboi:
> Agreed; though this now means quite a refactor of broker usage is required.

learning by doing :p
on a more serious note, i think we should do this because otherwise we are locked to 1.12

edmorley:
So I got this kind working after a refactor, however to then get the tests passing would have required duplication of main setup in test setup, or else yet more refactoring (eg to break out config from main).

Whilst many of these changes would likely be a good thing to do longer term, I'd like to avoid mixing too many changes into one PR (given the issues we had around eg the Metaas Go upgrade that touched hundreds of files).

As such I think the best way to solve this is just to remove support for the `-redisUrl` flag entirely, since it's not actually used, given that tests+production use the `REDIS_URL` env var instead.

edmorley:
Deployed to production

PR #211, opened 2019-10-29
Truncate the output if it reached size limit

dmathieu:
This is another approach at #209 which truncates the beginning of the content if the limit was reached.
Note that we will store the entire output in memory whenever there's a write happening (as opposed to using `APPEND` currently). So we will increase our memory footprint.


jabrown85:
Maybe we should consider storing a second redis key with the size written to busl and only do the tailing of the logs for the large ones? It would reduce the re-writing of large chunks of data to only the ones that exceed the limits in place. Also, we could not tail the logs but maybe halve the logs when the limit is reached so you only end up having to read from redis each time the limit is re-reached instead of every time. It potentially means dropping half of the logs for a single byte, but as long as we keep the tail I expect it's OK.

bensymonds:
> Maybe we should consider storing a second redis key with the size written to busl and only do the tailing of the logs for the large ones? It would reduce the re-writing of large chunks of data to only the ones that exceed the limits in place.

Yeah actually I noticed there is already a [`STRLEN`](https://redis.io/commands/strlen) command which would work for that, so that solves the problem for any streams which are below the limit (we'd only need to do that extra `STRLEN` call, and then if below the limit do the `APPEND`). But it still leaves the problem with the streams that are over the limit, and those are the ones that will disproportionally affect memory/latency so it didn't feel like a full solution.

> Also, we could not tail the logs but maybe halve the logs when the limit is reached so you only end up having to read from redis each time the limit is re-reached instead of every time. It potentially means dropping half of the logs for a single byte, but as long as we keep the tail I expect it's OK.

Interesting idea! Kind of like how some databases make extra space for data when a limit is reached - each time they allocate twice the space as they did before - just the inverse. That would definitely make things a lot better. Not sure if it would make me _fully_ comfortable with it though.

The only other thing I thought on this was whether we could use a different data structure in redis, one that lends itself to this access pattern of appending on the end and truncating at the start. Obviously this is more work but doesn't mean it should be ruled out. Eg if you use a list, you have things like https://redis.io/commands/ltrim, the docs for which funnily enough say:

> This is very useful when using Redis to store logs for example.

!

dmathieu:
LTRIM won't work here. It's a string we have, not a list. It seems the only way to trim a string on the server would be using [CHOP](https://github.com/RedisLabsModules/redex/pull/10/). But that's an unofficial module which isn't even merged into the repository.

Trimming half of the data if we're over the limit so we do it less often, and can still use APPEND works for me.

bensymonds:
> LTRIM won't work here. It's a string we have, not a list. 

Yes I said in my comment that to use LTRIM we'd have to use a different data structure, a list.

broker/io.go
```diff
@@ -50,13 +51,26 @@ func (w *writer) Write(p []byte) (int, error) {
...
+	if err != nil {
+		return 0, err
+	}
+
+	if l >= int(w.limit) {
+		c, err := redis.Bytes(conn.Do("GET", w.channel.id()))
+		if err != nil {
+			return 0, err
+		}
+		conn.Send("SET", w.channel.id(), c[len(c)/2:])
```
bensymonds: Just noting that the resulting length (after the truncation) will be somewhat unpredictable given `len(c)` will depend on how many bytes were appended when it went over the limit. Eg if the limit is 100, we're on 95 and 10 bytes are received, the length will be 105 in total which then gets truncated down to 52. I was wondering whether we could make the post-truncation length always be `w.limit/2` so it's predictable (in my example it would always be 50), but it doesn't seem like a big deal not to so no worries.

dmathieu: Write will usually send small chunks of data, both due to the nature of the stream (you rarely get 50MB of data at once), and because smaller byte arrays are much easier to handle.

PR #212, opened 2019-10-30
Store the output length in redis

dmathieu:
So we can truncate the content and still have the length incremented, to properly discard the content on a new request.


broker/io.go
```diff
@@ -283,8 +284,12 @@ func Len(wd io.WriteCloser) (int64, error) {
 	conn := redisPool.Get()
 	defer conn.Close()
 
-	strlen, err := redis.Int64(conn.Do("STRLEN", w.channel.id()))
+	strlen, err := redis.Int64(conn.Do("GET", w.channel.lengthID()))
```
freeformzSFDC: What happens when the key referred to by `lengthID()` isn't found?

jabrown85:
Good catch @bensymonds!

broker/io.go
dmathieu: We get a `redis.ErrNil`, which is handled below.

dmathieu: That's also already tested here: https://github.com/heroku/busl/blob/master/broker/io_test.go#L223

freeformzSFDC: Why not fall back to STRLEN on the channel id?

freeformzSFDC: FWIW: adding some code comments here that describe why things work the way they do would help future / new readers (doesn't have to be part of this PR of course).

dmathieu: I've added a method comment: 5a42568f26057e266dedcd13d3cf172931927a57

dmathieu:
You're right, the streams could have their output duplicated, as busl thinks there's nothing to discard. I've split this into 2 PRs.
This one now only stores the data. #213 uses it.

broker/io.go
```diff
@@ -285,6 +289,10 @@ func Len(wd io.WriteCloser) (int64, error) {
 
 	strlen, err := redis.Int64(conn.Do("STRLEN", w.channel.id()))
 	if err != nil {
+		if err == redis.ErrNil {
+			return 0, nil
+		}
+
```
bensymonds: Can you move this into #213 too? It's not required here.

dmathieu: Sure. I didn't do it because it doesn't matter. But it can go into the other commit too.

broker/io.go
```diff
@@ -275,6 +276,9 @@ func RenewExpiry(rd io.Reader) {
 }
 
 // Len returns the length of data already send to the reader
+// The length is stored in a separate key as opposed to using `STRLEN`, as the
+// content can be truncated if it reaches the size limit. When that happens, we
+// still want to increment the length though.
```
bensymonds: Ugh sorry I should have this needs moving too.

PR #213, opened 2019-10-31
Use stored length as the stream's length

dmathieu:
This splits #212 into 2 deployments to prevent streams with wrong output.


bensymonds:
How do we know when this one is safe to deploy?

dmathieu:
As soon as the app is fully restarted.
A restart will close all connections. On reconnection, we will start storing the length and we can then deploy this PR.

PR #215, opened 2019-10-31
Use stored len

dmathieu:
#213 was not merged against master.
Thanks @bensymonds


dmathieu:
cc @lillianzhang331 @SEAnderson07

PR #216, opened 2019-11-01
Log more info when `redis.Bool()` errors

bensymonds:
Thought I would have a go at this, for fun. Not really sure if I've done it right though.

We are getting some errors like:

```
redigo: unexpected type for Bool, got type string
```

(eg https://rollbar.com/Heroku-3/busl/items/221). Capture the reply so we can log it. Note `redis.Bool()` has a feature where if passed an error as second argument it will return it (so it can easily wrap `conn.Do()`). This is why there's no error return immediately after the `conn.Do()` call.

https://gus.lightning.force.com/a07B00000077cVqIAI


freeformzSFDC:
`redis.Bool` takes an `interface{}`.
`channel.id()` returns a `string`
The [current version of `redis.Bool` has no case to handle a `string`](https://github.com/gomodule/redigo/blob/4f569a4708584d192c9b2e49bb2d9936341f1e68/redis/reply.go#L201).

My suggestion is to either convert the `string` to a `[]byte` (ex: `exists, err := redis.Bool([]byte(reply), err)`) or PR redigo to add a case for string like so:

```diff
diff --git a/redis/reply.go b/redis/reply.go
index c2b3b2b..f45e181 100644
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -207,6 +207,8 @@ func Bool(reply interface{}, err error) (bool, error) {
                return reply != 0, nil
        case []byte:
                return strconv.ParseBool(string(reply))
+       case string:
+               return strconv.ParseBool(reply)
        case nil:
                return false, ErrNil
        case Error:
diff --git a/redis/reply_test.go b/redis/reply_test.go
index c7f9b28..bd38f99 100644
--- a/redis/reply_test.go
+++ b/redis/reply_test.go
@@ -121,6 +121,11 @@ var replyTests = []struct {
                ve(redis.Positions([]interface{}{[]interface{}{[]byte("1"), []byte("2")}, nil, []interface{}{[]byte("3"), []byte("4")}}, nil)),
                ve([]*[2]float64{{1.0, 2.0}, nil, {3.0, 4.0}}, nil),
        },
+       {
+               "Bool(string,nil)",
+               ve(redis.Bool("true", nil)),
+               ve(bool(true), nil),
+       },
 }
 
 func TestReply(t *testing.T) {
```

PS: Long term it's probably best to re-think the conversion of `[]byte` to `string` in `channel`'s methods.

freeformzSFDC:
Actually, I think I'm very wrong about all of that. :-(

freeformzSFDC:
Yep. I'm wrong, just ignore me.

freeformzSFDC:
PS: Log the channel.id() too as this may provide insight into what is happening. Essentially this could be a variation on someone trying to do the redis equivalent of SQL injection attack ?

bensymonds:
@freeformzSFDC Either way, thanks for the comments 😄

I think what must be happing is that redigo is getting some unexpected reply from the server. [`conn.readReply()`](https://github.com/gomodule/redigo/blob/4f569a4708584d192c9b2e49bb2d9936341f1e68/redis/conn.go#L534) is where the reply from redis is parsed and converted to the correct go type - you can see where it looks for a `:` and returns an `int64`. It's the `interface{}` from that function that ends up going into `Bool()`. That function must be getting something unexpected and return a `string`, which causes the error in `Bool()`. So my idea was just to log that string and see what it is 🤷‍♂

bensymonds:
> PS: Log the channel.id() too as this may provide insight into what is happening. Essentially this could be a variation on someone trying to do the redis equivalent of SQL injection attack ?

You can actually see the channel id in the rollbar occurrences (it's the `key`): https://rollbar.com/Heroku-3/busl/items/221. They all seem normal. Actually I forgot to mention - after opening this PR I learned that it appeared it was a single dyno that was having the issue. After the dyno was restarted the errors stopped. So that would imply it was some dyno/redis connection issue rather than user behaviour.

Interesting point about SQL-like injection attacks though - I hadn't understood fully but it does seems like we're taking the path from the URL and putting it straight into a redis command. I would have to investigate to see whether we're exposed there at all.

freeformzSFDC:
Interesting find re: dyno.

That is very strange.

bensymonds:
Thanks for the input everyone, but I think I'm going to close this. On further inspection I realise there is a lot I don't yet understand about the behaviour of busl and redis and the way errors are sent to rollbar, so it doesn't seem worth going with.

To me it looks like we need to take a bigger step back because there are _many_ errors related to redis (a lot more than just this one) - a brief visual scan down the [list in rollbar](https://rollbar.com/Heroku-3/busl/items/) shows that. But the first step would be to improve the way we send them to rollbar to improve the grouping because as it stands a lot of different errors end up being grouped together, because of the way the rollbar fingerprint is generated (for an example pick an item in rollbar and look at the messages in the occurrence list, eg https://rollbar.com/Heroku-3/busl/items/218/?item_page=0&item_count=100&#instances).

